PR #298, opened 2026-02-03
fix(audio): use direct ALSA hardware access instead of dmix

Lorti:
## tl;dr

I can now hear all the beeps and the funky music again 🕺

## Summary

Fixes audio playback issue in containers where dmix (ALSA's software mixing plugin) fails due to missing IPC/shared memory support.

**This PR supersedes and replaces PR #296**, which attempted to fix audio via group permissions but did not address the root cause.

## Problem

The audio service was logging repeated ALSA errors:
```
ALSA lib pcm_dmix.c:999:(snd_pcm_dmix_open) unable to open slave
Playback error: Unknown error 524 [default]
```

**Root cause**: ALSA's dmix plugin requires shared memory and Unix sockets for IPC, which don't work reliably in Docker containers. This is a fundamental incompatibility, not a permissions issue.

**Why PR #296 didn't work**: That PR added `group_add: ["29"]` for audio group access, but:
- The container runs as root (distroless default), which already has full device access
- group_add has no effect when running as root
- The real problem was dmix trying to use IPC, not file permissions

## Solution

Use direct hardware access (`hw:CARD,DEV`) instead of dmix:

1. **music_player.py**: Added `ALSA_DEVICE` environment variable support
   - Defaults to "default" for backward compatibility
   - Logs which device is being used for easier debugging

2. **docker-compose.hardware.yml**: Set `ALSA_DEVICE=hw:2,0`
   - Card 2 is the USB Audio Device on Raspberry Pi
   - Configurable via environment variable if hardware differs

3. **asound.conf**: Updated default ALSA config
   - Changed from dmix-based mixing to direct hardware access
   - Simplified config: `pcm.!default` → `hw:2,0`

## Test Plan

- [x] Verified USB audio device is card 2: `cat /proc/asound/cards`
- [x] Increased volume to audible level: `amixer -c 2 sset 'Speaker' 80%`
- [x] Tested direct hardware access on host: `speaker-test -D hw:2,0 -c 2`
- [x] Rebuilt and restarted audio service
- [x] Tested sound playback via gRPC: beep heard successfully ✅
- [x] Test music playback during game
- [x] Test sound effects during gameplay

## Notes

- Direct hardware access means only one application can use the audio device at a time
- This is acceptable for JoustMania since the audio service is the only audio source
- The dmix probe errors during miniaudio initialization are harmless (just backend detection)
- If your audio device is on a different card, adjust `ALSA_DEVICE` environment variable

## Related

- Supersedes PR #296 (audio group permissions) - that approach did not address the root cause
- Fixes audio regression introduced during microservices refactoring

### DIFF
--- a/services/audio/asound.conf
+++ b/services/audio/asound.conf
@@ -1,14 +1,17 @@
 # ALSA configuration for JoustMania container
-# Uses direct hardware access to avoid dmix/IPC issues in containers
+# Uses direct hardware access - simpler and more reliable in containers
 #
-# Override by mounting your own: -v /path/to/asound.conf:/etc/asound.conf:ro
+# Note: Only one audio stream can play at a time with this configuration.
+# The audio service handles this by managing playback channels.
 
-# Default output - direct hardware access
-# Card 2 is typically USB Audio on Raspberry Pi
-# Adjust if your audio device is on a different card
+# Default output - direct hardware access to USB Audio (Card 2)
 pcm.!default {
     type plug
-    slave.pcm "hw:2,0"
+    slave {
+        pcm "hw:2,0"
+        rate 44100
+        channels 2
+    }
 }
 
 ctl.!default {


aepfli:
To be honest, I never tried instructions - but I heard the congratulations etc.

PR #565, opened 2026-02-21
docs: Raspberry Pi audio setup (ALSA card + disable PipeWire)

Lorti:
## Summary

- Documents the `ALSA_CARD=2` requirement for Raspberry Pi deployments (cards 0 and 1 are HDMI outputs — unavailable without a connected display)
- Explains why PipeWire/WirePlumber must be masked on the host (it holds exclusive control of the USB audio device, blocking the Docker container's direct ALSA access)
- Includes commands to re-enable WirePlumber if needed later

## Test plan

- [ ] Documentation is accurate and commands work on a fresh Pi setup

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -827,3 +827,21 @@ grpcurl -plaintext localhost:4317 list
 ---
 
 Happy coding! 🎮
+
+## Raspberry Pi Audio Setup
+
+The audio service needs direct ALSA access to the USB audio device (card 2). The host's PipeWire/WirePlumber stack conflicts with this and must be masked:
+
+```bash
+systemctl --user mask wireplumber pipewire pipewire-pulse pipewire.socket pipewire-pulse.socket
+systemctl --user stop wireplumber pipewire.socket pipewire-pulse.socket
+```
+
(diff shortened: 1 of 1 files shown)


aepfli:
Questioning whether this is needed:

1. **`ALSA_CARD=auto` already handles this** — the audio service auto-detects the correct sound card at startup (see `alsa_config.py`). Hardcoding `ALSA_CARD=2` is fragile since card numbers can change between boots or when USB devices are plugged in different order. The existing `auto` default is the right approach for most users.

2. **PipeWire masking is host-specific ops knowledge, not development docs** — this is a deployment/ops concern for one specific Pi configuration, not something developers need. It also depends on the distro (some Pi images don't even run PipeWire). If we document it, it probably belongs in a deployment/operations guide, not DEVELOPMENT.md which is about building and running the codebase.

3. **The comment in docker-compose.yml already covers the override** — `ALSA_CARD=auto  # Auto-detect sound card; set to card number (e.g. "1") to override` tells users exactly what to do if auto-detect doesn't work for their setup.

If this came up as a real issue on your Pi, maybe the auto-detection logic in `alsa_config.py` needs improving instead? That would fix it for everyone rather than documenting a manual workaround.

Issue #779, opened 2026-06-05
Integration tests: concurrent games + shadow-game isolation


aepfli:
Part of #774 (shadow games). Phase 5 of 5 — the payoff for CI. Related: #770 (integration test speed).

## What parallel-game support unlocks

- **Concurrent lifecycle tests** against one compose stack: each test starts its own headless game with its own mock controller set and game_id — no shared global game. Enables pytest-xdist for non-menu tests (start with 2 workers given shared compose resources).
- **Per-game cleanup**: split the global `ensure_game_stopped` fixture into a menu-scoped one (sequential menu-flow tests, force-end primary) and a per-test game_id-scoped cleanup (force-end by id + remove reserved controllers).
- **Shadow-game isolation regression test**: start a menu-flow game with unreserved mocks AND a headless game with reserved mocks simultaneously; assert the menu lobby never sees the reserved controllers and its ready-count is unaffected. This is the direct regression test for the whole #774 feature.

## Helper changes (`tests/integration/helpers.py`, `conftest.py`)

- `start_game_headless(start_config) -> game_id` using `StreamGameEvents` directly
- `GameEventCollector` becomes game_id-aware (filter by `GameEvent.game_id`)
- `setup_mock_controllers(reserved=..., tag=...)`

## Constraint

Menu-flow tests remain sequential — the menu is single-lobby and that does not change.

aepfli:
Done in PR #827: start_game_headless + game_id-aware GameEventCollector, reserved controller setup/cleanup helpers, GAME_MAX_CONCURRENT_GAMES=4 in CI compose, and three new tests (concurrent two-game isolation, shadow-alongside-menu — the #774 regression net — and natural-end-then-restart for the #793 CI bug).

Issue #777, opened 2026-06-05
Controller manager: reserved mock controllers (hidden from menu)


aepfli:
Part of #774 (shadow games). Phase 3 of 5. Scheduled after #730.

## Problem

The menu learns about controllers exclusively via button-stream connect events (`controller_manager/servicer.py:155-181`) plus the `connected_serials` roster in each ButtonEvent (:172). If the agent adds mock controllers for a shadow game, the menu would show them in the lobby, write LEDs to them (last-write-wins, no arbitration), and count their trigger presses toward ready state.

A serial-prefix convention can't work: integration tests legitimately use unreserved mock controllers through the menu flow, so the prefix can't distinguish "agent-reserved mock" from "test's menu-flow mock".

## Approach: explicit reservation flag + announce-suppression

- `proto/controller_manager_mock.proto`: add `bool reserved` + `string tag` to `AddControllerRequest` and `AddControllersRequest` (tag identifies the owning agent/game — aids cleanup of orphans and debugging). Backward compatible (default false).
- `multiplexer/mock_adapter.py`: store `reserved`/`tag` in controller state; plumb through `mock_control_service.py`.
- `controller_manager/servicer.py`: suppress reserved controllers from **both leak paths** — button-stream connect events (initial snapshot AND live connect path) and the `connected_serials` roster.
- Reserved controllers stay **fully usable**: gameplay-data streams (with serial filter) and explicit-serial LED writes still work — they are just never announced to button-stream consumers.

## Verification

Add 2 reserved + 2 unreserved mocks; assert a `StreamButtonEvents` subscriber sees only the 2 unreserved (connect events and roster); assert `StreamGameplayData` filtered to a reserved serial still streams motion; assert LED set by serial works on a reserved controller. Existing menu integration tests unaffected.

aepfli:
Done in PR #784: reserved/tag on mock AddController(s), suppression of connect events and connected_serials roster for reserved controllers; gameplay-data streaming and LED-by-serial unaffected.

Issue #775, opened 2026-06-05
Game coordinator: multi-session refactor (dict[game_id, GameSession])


aepfli:
Part of #774 (shadow games). Phase 1 of 5. Scheduled after #730.

## Goal

Replace the single-game state machine in `services/game_coordinator/servicer.py` with a `dict[game_id, GameSession]` so multiple games can run concurrently.

## Approach

1. **Characterization tests first** — pin current single-game behavior at the servicer level: start via `StreamGameEvents(start_config)`, `GetGameState`, `ForceEndGame`, "Game already in progress" rejection, event ordering, metrics side-effects.
2. **`GameSession`** (new `services/game_coordinator/game_session.py`) bundling `game_id`, `game_name`, `players`, `game_config`, `start_time`, `state`, `current_game`, `event_bus`, `thread`, `running`, `parent_context`. Each game already runs its own thread + asyncio loop + `GrpcClientManager` (created inside the thread's loop), so this is a state-container move, not a concurrency redesign.
3. **Per-session EventBus** with a persistent **primary bus** owned by the servicer: the menu-driven game uses it, so today's zero-arg `StreamGameEvents` subscribe semantics are preserved exactly — menu needs zero changes. Agent/shadow sessions get their own bus.
4. **Configurable max-concurrent-games cap** (flagd or env, small default).

## Load-bearing bugs to fix (required for parallelism)

- `game_id = f"game_{int(time.time())}"` (servicer.py:115) collides at second resolution → uuid-based id.
- `clear_all_player_analytics()` (metrics.py:258-282) does a global `_metrics.clear()` — the first game to end wipes the live game's player gauges. Use the existing targeted `clear_player_analytics(serial, game_id)` (metrics.py:231) keyed off the ending session's player set.

## Metrics

Add a `game_kind` label (`primary`/`shadow`) to lifecycle metrics (`active_game`, `active_players`, `games_started_total`, `game_duration_seconds`, ...). Note: Grafana queries that `sum()` over game metrics need a `game_kind` filter review when this lands.

## Verification

Characterization tests pass unchanged after refactor; new test: two concurrent `start_config` streams yield distinct game_ids and independent event streams.

aepfli:
[...]

Issue #774, opened 2026-06-05
Shadow games: mock-only parallel games for the agent


aepfli:
## Summary

For the agent ACT layer, the agent should be able to start games using only mock controllers — possibly several in parallel — as shadow/evaluation games running alongside a real game with real players.

Analysis doc: `docs/research/774-shadow-games.md` (PR to follow).

## Key findings

Most of the infrastructure already exists:
- Full mock controller infra (`MockControllerService` on :50062, mock adapter always injected alongside hardware adapters)
- Headless game start via `StreamGameEvents(start_config)` — no menu needed
- `StreamGameplayData` already supports multiple concurrent subscribers with per-stream serial filters

The actual blockers:
1. Single-game state machine in game_coordinator (plus two load-bearing bugs for parallelism: second-resolution `game_id` collision, and `clear_all_player_analytics()` global wipe)
2. No `game_id` routing in `game_coordinator.proto`
3. Menu interference: agent-added mock controllers would appear in the lobby, get LED writes, and count toward ready state — no reservation concept
4. Agent service (Go) is observation-only today

## Phases

- [x] #775 — Game coordinator: multi-session refactor (`dict[game_id, GameSession]`)
- [x] #776 — Proto: `game_id` routing + `ListGames` RPC
- [x] #777 — Controller manager: reserved mock controllers (hidden from menu)
- [x] #778 — Agent: shadow-game runner (Go) — depends on #730
- [x] #779 — Integration tests: concurrent games + shadow-game isolation (related: #770)
- [x] #826 — Parallelize game-mode lifecycle tests via headless shadow games (follow-up, serves #770)
- [x] #837 — Shadow game governance: origin marking + real-game resource gate
- [x] #838 — Per-game flag evaluation: gameId in EvaluationContext (interventions + calibration)

Sequencing: all implementation scheduled **after #730** (intervention API); the Go shadow-game runner reuses #730's client plumbing.

## Integration test benefit

Concurrent lifecycle tests against one compose stack (helps #770), per-game_id cleanup instead of the global force-end fixture, and a direct "shadow game alongside menu game" regression test. Menu-flow tests stay sequential (menu remains single-lobby).

aepfli:
**M6 Shadow Games: all phases complete.** 🏁

| Phase | Delivered |
|---|---|
| #775 multi-session coordinator | PR #793 — GameSession registry, primary bus, uuid game_ids |
[...]

PR #827, opened 2026-06-05
test: concurrent headless games + shadow-vs-menu isolation (#779)

aepfli:
## What

Phase 5 of #774 (shadow games): integration coverage for concurrent headless games and shadow-vs-menu isolation, built on the multi-session game coordinator (#775/#776, PR #793) and reserved mock controllers (#777, PR #784). Diff is limited to `tests/integration/` + `docker-compose.ci.yml`.

## Helper API (`tests/integration/helpers.py`)

- `setup_mock_controllers(..., reserved=False, tag="")` — passes through to the `AddControllers` reservation fields. Reserved/tagged requests always add fresh controllers (so the reservation applies); the default unreserved path keeps the existing reuse fast-path.
- `start_game_headless(game_client, start_config) -> (game_id, collector)` — starts a game directly via `StreamGameEvents(start_config)` (no menu) and returns the coordinator-assigned `game_id` plus a running collector. `build_start_config(mode, serials)` builds a minimal per-mode `StartGameConfig`.
- `GameEventCollector` is now game_id-aware: an explicit `game_id` subscribes to that session's stream and filters to it; a `start_config` collector learns and filters by the assigned id. **A plain zero-arg primary collector stays unfiltered** (see note below).
- `force_end_game_by_id(game_id)`, `list_games()`, `remove_reserved_controllers(tag)`, `list_mock_controllers_detailed()`.
- `ButtonStreamObserver` — subscribes to `StreamButtonEvents` (the exact source the menu reads) to assert reserved controllers never appear in any connect event or `connected_serials` roster.
- `get_player_states` / `kill_player_verified` / `kill_players_until_one_remains` accept an optional `game_id` for per-session kills.

## Fixtures (`tests/integration/conftest.py`)

- `ensure_game_stopped` unchanged (menu-flow tests, force-ends the primary).
- New `headless_cleanup` factory: each headless test registers its owned `game_id`s and reserved tags; teardown force-ends each game and sweeps each tag, even on failure.

## CI cap (`docker-compose.ci.yml`)

- `GAME_MAX_CONCURRENT_GAMES=4` on the game-coordinator service. **CI only** — the coordinator's production default stays 1. The override lives in CI compose, not baked into images.

## New tests (`tests/integration/test_concurrent_games.py`)

- **`test_two_concurrent_headless_games`** — two headless games (FFA + Teams) on distinct reserved controller sets: distinct game_ids, per-collector event isolation, `ListGames` shows both RUNNING, `GetGameState(game_id)` returns each roster, ending one leaves the other running with an uncontaminated stream.
- **`test_shadow_game_isolated_from_menu`** — the direct regression net for #774: a menu-flow game on 4 unreserved mocks AND a headless shadow game on reserved mocks at the same time. Asserts reserved controllers never leak into the menu's button-stream view, shadow events carry the shadow game_id and never reach the primary stream, ending the shadow leaves the menu game running, and the menu restores lobby colors on its own controllers.
- **`test_headless_natural_end_then_restart`** — runs a headless game to its natural win condition, then immediately starts another. This is the regression for the #793 CI incident, where an ENDED session pinned the primary slot and the next start was rejected with "Game already in progress" (the coordinator now lazily retires ENDED sessions on the next start).

## Note: zero-arg collector must stay unfiltered

While bringing up the suite, the game_id-aware collector initially auto-locked its filter onto the first event carrying a game_id. For a zero-arg primary collector that subscribes before the game starts, that first event can be a stale/late event from a previously-ended game — the collector then dropped the new game's `game_started`, breaking the existing menu-flow and intervention tests. Fixed by only enabling game_id filtering for collectors explicitly bound to a session (explicit `game_id` or `start_config`); plain zero-arg primary collectors collect everything, exactly as before.

## Verification

Full integration suite (`make test` equivalent, prebuilt images): **21 passed**. Includes all 13 menu-flow lifecycle modes, all intervention-flow tests, and the 3 new tests.

## Follow-up

pytest-xdist adoption was deliberately deferred. The headless data plane multiplexes (so parallel non-menu tests are viable), but the menu is single-lobby, so menu-flow tests must stay sequential. Running sequentially within one stack is fine for now; xdist with worker-pinned menu tests is a clean follow-up if CI time becomes a concern (related: #770).

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -55,6 +55,11 @@ services:
     environment: !override
       OTEL_SERVICE_NAME: game-coordinator-service
       OTEL_EXPORTER_OTLP_ENDPOINT: http://otel-collector:4317
+      # Raise the concurrent-games cap for the integration suite only (#779).
+      # The coordinator defaults to 1 (production default, unchanged); the
+      # concurrent/shadow-isolation tests need multiple live sessions on one
+      # stack. This override stays in CI compose; production keeps cap=1.
+      GAME_MAX_CONCURRENT_GAMES: "4"
 
   # Pairing Daemon - Requires hardware, excluded from CI
   pairing-daemon:
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -116,6 +116,7 @@ async def ensure_game_stopped(docker_compose):
     Use this fixture explicitly in tests that start games to prevent
     'Game already in progress' errors between tests.
     """
+
     async def force_end_game():
         """Force end any running game."""
         try:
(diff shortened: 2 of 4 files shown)


aepfli:
**Local verification notes**

Full integration suite (prebuilt images): first run **21/21 passed**; a second run showed **20 passed, 1 failed** where the single failure was `test_full_game_lifecycle[FightClub]` timing out on the menu game start (`game_started` not emitted).

This FightClub failure is a **pre-existing menu-flow flake (#757), not introduced by this PR**. Verified by running `test_full_game_lifecycle[FightClub]` in isolation against the PR base commit (`cc9a74b7`, none of this PR's changes) — it fails there identically (`Collected 0 events`, `game_started` timeout). It is order-sensitive: it surfaces when FightClub runs first/alone with freshly-added mock controllers, where the menu occasionally hasn't registered the controllers before the ready-up triggers fire.

All three new concurrent/isolation tests and all four `test_intervention_flow` tests pass reliably across runs.

Issue #826, opened 2026-06-05
Parallelize game-mode lifecycle tests via headless shadow games


aepfli:
Follow-up to #779, part of #774 (M6). Directly serves #770 (integration CI speed).

## Idea

Now that the coordinator supports concurrent games (#775/#776) and reserved mock controllers are invisible to the menu (#777), the 14 per-mode lifecycle tests can run as **headless shadow games, 4 at a time**, instead of strictly sequentially through the menu.

## Design constraints (learned in M6)

- **Reserved controllers are mandatory**, not optional: the menu service writes lobby LEDs to every controller it can see; parallel unreserved games would race the LED assertions.
- **`asyncio.gather` batching inside one parametrized test, NOT pytest-xdist**: the `docker_compose` fixture is session-scoped, but each xdist worker gets its own session → 4 compose stacks. In-process async batches share the one stack.
- **Batch by expected duration** (wall time = Σ of per-batch maxima): group the slow modes (Tournament, FightClub) together rather than letting each anchor a separate batch.
- **Keep a thin sequential menu-flow set** (2–3 modes via `start_game_via_menu`): lobby ready-up, mode selection, and lobby-color-restore are menu behavior and the menu remains single-lobby. Parallel headless tests cover per-mode game logic (colors via GetColor per serial, deaths, win conditions, end events) but skip lobby-restore checks.
- Cap: `GAME_MAX_CONCURRENT_GAMES` in `docker-compose.ci.yml` must cover batch size + a possible menu game (#779 sets it to 4; bump to 5 if a menu game runs concurrently).

## Builds on (#779 helpers)

`start_game_headless()`, game_id-aware `GameEventCollector`, `setup_mock_controllers(reserved=, tag=)`, per-game_id cleanup.

## Expected gain

~3–4.5 min of sequential game time compresses to ~1–1.5 min → roughly 2–3 min off the integration job. Compose build/up/teardown unchanged (that part is #767/#770 territory).

aepfli:
Done in PR #834: per-mode lifecycle coverage moved to headless shadow-game batches (asyncio.gather, duration-grouped, reserved controllers). Measured: Integration Tests job 9m50s → 5m4s (~48% faster). Plus the suite-wide coordinator_idle autouse fixture enforcing the no-leftover-games invariant with loud leak reporting.

Issue #770, opened 2026-06-05
Speed up Integration Tests CI job (currently ~9 min, ~5 min after #767)


aepfli:
## Context

The `Integration Tests` job gates every PR. Before #767 it ran ~9 minutes; #767 cut the test phase from 515s → ~200-230s (verified kills no longer over-wait), so the job should now land around ~5 min. There is room to get to ~3 min.

Timing breakdown from run [26967743303](https://github.com/WatchMeJoustMyFlags/JoustMania/actions/runs/26967743303) (GitHub Actions overhead is only ~16s — everything else is inside the pytest session):

| Chunk | Time | Notes |
|---|---|---|
| Image pull + `compose up --wait` + venv | ~90s | inside the session-scoped fixture |
| FightClub | 34s | 6 rounds × 3.5s (1s pause + 2s invincibility + buffer) |
| Tournament | 20s | 3 rounds × 4.2s invincibility wait |
| Werewolf / Zombies / Traitor | 14–17s each | |
| Other 5 modes | 7–13s each | |
| Per-test fixed cost | ~1.2s × 10 | menu stop/start sleeps, force-end sleeps |

**Scope note:** the full stack (including the observability services) stays up in CI — we want the whole stack exercised. No compose slimming.

### 1. Shard modes across parallel matrix jobs (biggest wall-clock win)
Tests are inherently sequential against one stack (one game at a time), but a 2–3 way matrix (`pytest -k "JoustFFA or JoustTeams or ..."`) runs concurrent stacks on separate runners.
Wall clock ≈ stack startup (~90s) + slowest shard (~70s) ≈ **3–4 min**, at the cost of 2–3× runner-minutes. Balance shards so the slow modes (FightClub, Tournament) land in different jobs.

### 2. Faster CI game timings (~30–40s test-phase win)
- `invincibility_seconds`: CI currently uses `2s` (already the smallest variant) — add a `0.5s` variant for `game.ci.json` → FightClub round wait 3.5s→2.0s, Tournament 4.2s→1.5s
- `fight_club.min_rounds`: 5 → 2 in CI (test then runs 3 rounds instead of 6)
- `winner_rainbow_duration_ms`: add a 250ms CI variant (currently 1000ms)
- The hardcoded 1s "show winner" sleep in `base.py` (`_game_loop`, "Win condition met, keeping game active for 1 second") could come from runtime config
[...]

Issue #757, opened 2026-06-05
Flaky integration test: test_full_game_lifecycle times out waiting for game_ended


aepfli:
## Problem

`tests/integration/test_full_game_lifecycle.py::test_full_game_lifecycle` intermittently fails in CI with:

```
TimeoutError: Game did not emit any of ['game_ended', 'game_force_ended', 'game_error'] within 15 seconds
```

It is **mode-roulette** — observed failing on different parametrized modes across runs:

- Run [26964884842](https://github.com/WatchMeJoustMyFlags/JoustMania/actions/runs/26964884842) (PR #748): JoustFFA, JoustTeams, Zombies (+ "LED stuck at death effect" assertions)
- Run [26964848995](https://github.com/WatchMeJoustMyFlags/JoustMania/actions/runs/26964848995) (main, no agent code — proves it's pre-existing): identical failures
- Run [26967743303](https://github.com/WatchMeJoustMyFlags/JoustMania/actions/runs/26967743303) (main, #749 merge): `[Swapper]` — 1 failed, 9 passed → **main is currently red because of this**

## Why it matters

Every PR merge has a coin-flip chance of a red `Integration Tests` job unrelated to its changes (it cost PR #748 a red X and currently paints main red). During PR #748's investigation the "collector/agent exporter" hypothesis was empirically disproven — the flake reproduces on main without any agent code.

## Hypotheses to investigate

- High-death-count modes time out at 15s under CI load (mock backend + full docker stack on a shared runner) — the timeout may simply be too tight
- Death-effect LED assertions racing the game-end transition ("LED stuck at death effect")
- Game loop frame drops under CI CPU contention delaying the win-condition check

## Acceptance

- [ ] Root cause identified (or timeout/assertion hardened with justification)
- [ ] `test_full_game_lifecycle` green across 10+ consecutive CI runs

PR #834, opened 2026-06-05
test: parallelize game-mode lifecycle coverage via headless shadow-game batches (#826)

aepfli:
## Summary

Implements #826 (M6 follow-up, serves #770): per-mode game-logic coverage moves off the sequential menu flow into **headless shadow games running 4 at a time** via `asyncio.gather` on the single shared compose stack.

### Design (per the issue)
- **Duration-grouped batches**: slow modes (Tournament, FightClub, Werewolf) pooled into one batch with Zombies filling the 4th slot; fast modes in the other. Wall time per batch = its slowest mode, so 12 sequential mode-runs become 2 batch-runs.
- **Reserved controllers per mode** (distinct tags) — mandatory, since the menu writes lobby LEDs to any controller it can see.
- **`asyncio.gather`, NOT pytest-xdist**: the `docker_compose` fixture is session-scoped; xdist workers would each boot their own stack.
- **Thin menu-flow set retained**: `test_full_game_lifecycle` keeps JoustFFA + JoustTeams for menu ready-up / lobby-color-restore coverage (single-lobby, inherently sequential). FightClub leaves the menu-start path entirely — which should also sidestep the #757 menu-start flake.
- Failures surface per mode: `gather(return_exceptions=True)` + mode-tagged re-raise, double cleanup (per-coroutine `finally` + `headless_cleanup` tag sweep).

### Timing
Local before/after timing was not completed (the implementing agent stalled on the baseline run; see below). The honest comparison is CI's Integration Tests job: **baseline 9m50s** on #827's run — compare against this PR's job time.

### Note on provenance
The implementation was committed by a background agent that then stalled for ~4h waiting on a background baseline-suite run that never started. The code was reviewed, rebased onto current main (clean, including #824), and linted; the integration suite run on this PR's CI is the primary verification. If CI shows mode failures under 4-way parallelism, that itself is valuable signal (product race vs test design) — do not merge until Integration Tests is green.

Stacked-on-#827 originally; rebased onto main after #827 merged.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -212,6 +212,35 @@ async def ensure_game_stopped(docker_compose):
     await force_end_game()
 
 
+@pytest.fixture(autouse=True)
+async def coordinator_idle(docker_compose):
+    """Enforce the no-leftover-games invariant before EVERY integration test.
+
+    Tests must leave the coordinator idle. A live session leaking out of a
+    test occupies one of the GAME_MAX_CONCURRENT_GAMES slots, so the next test
+    that starts a game gets "Game already in progress" (this rejected one mode
+    per batch on the first CI run of the parallel lifecycle tests). This
+    fixture force-ends any leak and reports it loudly — when it fires, the
+    LEAKING test is the one that ran immediately before the reporting test.
+
+    NOTE (cross-test parallelism): this global quiesce assumes the sequential
+    single-process suite. If pytest-xdist with a shared stack is introduced,
+    scope this to tests needing exclusive coordinator access instead.
+    """
+    from tests.integration.helpers import quiesce_coordinator
+
+    host = docker_compose.get_service_host("game-coordinator", 50053)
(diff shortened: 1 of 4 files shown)


aepfli:
**Measured result — Integration Tests job:**

| | time |
|---|---|
| Baseline (#827, sequential modes) | **9m50s** |
| This PR (parallel batches) | **5m4s** |

**~48% faster** — beats the 2–3 min estimate from #826.

Hardening added after the first runs (each found by CI, all test-side, no product bugs):
1. Autouse `coordinator_idle` fixture — enforces the no-leftover-games invariant before every test and loudly names any leaked session (the leaker is the test right before it in the log)
2. Start retry for transient cap rejections (sibling games in their ENDING window)
3. Polling LED check instead of one-shot (feedback stream lags under 4-way load)
4. RandomTeams `pre_end_delay=6.5s` — its 5s team-formation phase swallows kills (`TEAM_FORMATION_DURATION`, random_teams.py)

Issue #837, opened 2026-06-05
Shadow game governance: origin marking + real-game resource gate


aepfli:
Follow-up to #774. Design gap found while wiring agent experiments to interventions (#730).

## Problem (current behavior)

"Primary" is **first-come-first-served**: whoever starts a game while no primary is active claims the primary slot (`_start_game_from_config`). Consequences:
- An agent game started during idle time becomes primary → blocks real players at the prod cap of 1 ("Game already in progress" at the lobby).
- With cap > 1 it inverts: the agent's game stays primary and a later REAL menu game is classified `shadow` — wrong `game_kind` metrics, and interventions (primary-only via `get_live_game()`) target the agent's game instead of the real one.

## Decisions (user, 2026-06-05)

1. **Origin marking**: the menu marks its starts as the real game (e.g. `claim_primary`/origin field on `StartGameConfig`); agent/headless games are ALWAYS shadow. Primary is reserved for the real game.
2. **Resource gate**: shadow games must not run while a real game is live — the real game gets all performance/resources.
   - Shadow start rejected while a real game is STARTING/RUNNING
   - A real game starting **preempts** (force-ends) any running shadow games
3. **Flag-configurable policy** (e.g. `game.shadow_policy`: `block` prod default / `allow` for CI) — the #779 shadow-vs-menu isolation test deliberately runs both concurrently and needs `allow`.

## Notes

- Agent experiment flow becomes: idle time only → always shadow → evicted on real-game start. Combined with gameId flag context (see companion issue) this gives safe idle-time experimentation.
- Touches: `game_coordinator/servicer.py` (origin, gate, preemption), `proto/game_coordinator.proto` (origin field, additive), `menu/servicer.py` (set origin), flagd game domain, CI compose/test updates.

Issue #766, opened 2026-06-05
Calibration surface: promote audited constants to flags (#722 §2b/§11)


aepfli:
## Context

A codebase-wide constant audit (PR #765) extended the #722 intervention surface research with a **calibration surface**: hardcoded constants upstream of the intervention pipeline that bound what the M2 intervention API (#730) can achieve. The intervention plumbing is done (applier in `services/game_coordinator/interventions.py`, Go flags package in `services/agent/flags/`); these insights are not yet implemented.

See [722-intervention-surface.md §2b and §11](../blob/main/docs/research/722-intervention-surface.md) (placement rule, tiers, follow-up plan F1–F8).

### F1 — Death/warning threshold tables → `game.json`
- Add object-typed flag `thresholds` (the four 5-element arrays `SLOW_WARNING`/`SLOW_MAX`/`FAST_WARNING`/`FAST_MAX`) with current values (`games/base.py:95-98`) as defaults; per-mode `zombie.thresholds` / `werewolf.thresholds` overrides (`zombie.py:41-47`, `werewolf.py:34-40`)
- Read once in `BaseGameMode.__init__` (init-frozen by design); replace module-constant uses at `base.py:878-879, 1671-1672`
- Validate shape (5 entries, warning < max per row); fall back to defaults on malformed values

### F2 — Durations, roles, scoring → `game.json`
- `death_grace_period_seconds` (base.py:106), `nonstop.respawn_seconds` / `nonstop.spawn_protection_seconds` / scoring weights (nonstop_joust.py:32-33,494), `fight_club.round_seconds` (fight_club.py:29), `tournament.match_seconds` / `tournament.time_between_matches_seconds` (tournament.py:40-42), `werewolf.werewolf_fraction` (werewolf.py:117), `traitor.count_tiers` (traitor.py:110-116), `zombie.initial_count` / respawn min/max (zombie.py:32-34)
- Each mode reads at init via the game-domain client; **no proto changes** — calibration flags bypass `StartGameConfig`

### F3 — Music schedule windows → `game.json` (depends on F2)
- `music.windows` object flag covering `MIN/MAX_MUSIC_FAST/SLOW_TIME` + end-game variants (base.py:72-81); consumed by `_music_loop` (base.py:1750-1754)
- Define the `pacing_profile` presets (`calm`/`default`/`frantic`) over these windows for F6

### F4 — Perception thresholds → `agent.json`
[...]

Issue #730, opened 2026-06-04
JoustMania intervention API (informed by intervention surface research)


aepfli:
## What to build

The **JoustMania intervention API**: the surface through which the agent's dispatched actions actually take effect in the game. This is the `dispatch(action)` target of the decision loop.

**Directly informed by the intervention surface analysis in #722** — the set of interventions exposed here must match the `interventions.allowed` values proposed by that research. Do not implement interventions beyond what #722 identifies as possible and meaningful.

## Scope

- API in JoustMania exposing the interventions identified in #722
- Each intervention addressable by the action `type` the agent produces, so `dispatch(action)` maps 1:1 to API calls
- Interventions operate within the physical-game model: movement thresholds, eliminations, and session dynamics driven by accelerometer data — no kill relationships, no virtual health
- Intervention effects observable in the game's OTel signals so fitness functions can evaluate their impact

## Acceptance Criteria

- [ ] Intervention API implemented in JoustMania covering the interventions from #722
- [ ] Agent `dispatch(action)` invokes the API and the game state changes accordingly
- [ ] Action types align exactly with the `interventions.allowed` flag values
- [ ] Intervention effects show up in the observable per-player/per-session signals
- [ ] Interventions blocked by the permission layer never reach the API

PR #890, opened 2026-06-05
feat: gameId evaluation-context routing for interventions + calibration (#838)

### LINKED ISSUE #838: Per-game flag evaluation: gameId in EvaluationContext for interventions + calibration
Follow-up to #774, extends #730's intervention control plane. Companion to the shadow-game governance issue.

## Decision (user, 2026-06-05)

Use OpenFeature for agent experiments by adding **gameId to the EvaluationContext**, so the existing flag config mechanism also drives testing/optimization runs on shadow games.

## Design

- #730 already evaluates interventions with `targetingKey=serial` for per-player rules. Add `gameId` as a context attribute.
- `InterventionManager` evaluates per **live session** (iterate the session registry) instead of primary-only `get_live_game()`, building context `{gameId, targetingKey=serial}` and applying results to that session's game.
- Flag rules can then target a specific game: `{"if": [{"==": [{"var": "gameId"}, "game_abc"]}, "experiment_variant"]}` — production games hit the default (no-op) variant and are untouched.
- **Calibration flags too** (#766): flags read at game init (threshold tables, pacing) evaluated with gameId context → different shadow games get different calibration variants = real A/B experiments via flag config alone.
- The agent's experiment loop: start shadow game (idle time only, per governance gate) → write gameId-scoped flag rules → observe via OTel/game events → compare variants.

## Constraints

- Default variants MUST be behavior-neutral (same rule as #766 calibration promotion).
- Per-player targeting (serial) and per-game targeting (gameId) compose in one rule.
- Interventions on the real game must be provably unaffected — needs a regression test: gameId-scoped intervention flag does not touch a concurrently running real game (builds on #779 isolation harness, requires shadow_policy=allow in CI).

aepfli:
Closes #838. Follow-up to #774; builds on #730 (intervention control plane), #793 (multi-session + game_id routing), and #837/#839 (shadow-game governance). Adds `gameId` as an OpenFeature EvaluationContext dimension so flag config alone can scope interventions AND calibration to a specific game — enabling idle-time agent A/B experiments on shadow games without touching production games.

## Design decision: un-targeted = primary-only
Interventions are now evaluated **per live session** instead of primary-only. The `InterventionManager` gains a `get_sessions` seam returning one `SessionView` (game_id, game, publish) per live game (primary + every shadow). Each intervention flag is evaluated against each session with a per-session `{gameId, targetingKey=serial}` context, applied to that session's game, and published on **that session's** EventBus.

**Backward compatibility is the load-bearing property.** A flag with no `gameId` targeting rule resolves to its default (no-op) for shadow games and to the configured value for the primary, so it touches **only the primary** — exactly today's semantics (least surprise; #780's e2e tests stay semantically untouched). Shadow games only receive interventions **explicitly** targeted at their `gameId`. Default variants remain behavior-neutral, so production games with no targeting rule are untouched.

## Seam chosen
- `get_game` → `get_sessions`: the manager iterates a `SessionView` list. The legacy `get_game` callback still works, synthesizing a single primary-only `SessionView` with an empty `game_id` (no `gameId` added to the context) and the primary publisher — so the enforcement chain, handlers, and #780 e2e flow are unchanged. No restructuring of the enforcement chain.
- Change-detection state (edge nonce / state value) is namespaced **per game_id** (`dict[game_id, dict[flag_key, value]]`) with stale-bucket pruning as shadows come and go.
- `InterventionContext` gains `game_id` so per-player handlers (`player_sensitivity_factor`, `shield_seconds`) re-resolve `resolve_player_targets(game_id=...)` — letting per-player (`targetingKey`) and per-game (`gameId`) rules **compose in one rule**.
- Battery guard now searches every live session for the serial, so a gameId-targeted player intervention is gated against its actual game.

## Calibration (#766)
The init-frozen `read_object/float/int_flag` helpers gain an optional `game_id` threaded into their EvaluationContext (via `_calibration_context`). `game_id` is plumbed from `self.game_id` at game construction (`base.py` + every mode subclass: zombie/werewolf/tournament/nonstop/traitor/fight_club). Empty/None adds no `gameId` attribute — a pure context addition, **identical resolution to today** unless a targeting rule references `gameId`. flagd targeting can now vary threshold tables / pacing / durations per game = real A/B shadow experiments via flag config alone.

## Compatibility argument
- All 919 pre-existing game_coordinator tests + 355 lib tests stay green (the only test edits adapt white-box callers of the changed internal `_enforce_and_dispatch`/`_evaluate_one` signatures and read-helper mock lambdas — no behavioral assertions changed).
- Default flagd configs are untouched (behavior-neutral); no schema change was required — the agent writes `gameId` targeting rules dynamically, exactly as it already writes `targetingKey` rules.

## Test coverage
- **Unit** `services/game_coordinator/tests/test_gameid_routing.py`: gameId-targeted edge fires only for the matching session (primary untouched); un-targeted = primary-only; per-player + per-game compose and mutate real game state; per-session change detection independent; legacy `get_game` path adds no gameId.
- **Calibration** `lib/tests/test_feature_flags.py`: `_calibration_context` gameId presence; read helpers thread gameId; targeted variant for matching game vs baseline for others / un-targeted.
- **Integration regression** `tests/integration/test_gameid_intervention_isolation.py` (the issue's required test): with `shadow_policy=allow` (CI default), a real menu game + shadow game run concurrently; a gameId-scoped `eliminate_player` flag targeting the SHADOW id eliminates the shadow victim while the real game's player stays alive and its stream sees no intervention; an un-targeted eliminate still hits the primary. All waits poll (no fixed-sleep one-shots).

## Scope
game_coordinator (`interventions.py` per-session seam, `servicer.py` `get_intervention_sessions` + battery search, difficulty/lifecycle handler `game_id` plumbing, factory/base + mode init reads) + `lib/feature_flags.py` context plumbing + tests. **Did NOT touch** the Go agent, menu, controller_manager, or proto. flagd configs unchanged (no behavior-neutral schema addition was needed).

## Follow-up
Writing `gameId`-scoped flags **from the Go agent** (the experiment-loop writer) is a deliberate follow-up — this PR makes the Python side context-aware and proves the end-to-end isolation with the file-writer the agent's ActionSink mirrors.

### DIFF
--- a/lib/feature_flags.py
+++ b/lib/feature_flags.py
@@ -178,7 +178,21 @@ def get_flag_client(domain: str):
     return api.get_client(domain=domain)
 
 
-def read_object_flag(domain: str, flag_key: str, default: dict) -> dict:
+def _calibration_context(game_id: str | None) -> EvaluationContext:
+    """Build the EvaluationContext for an init-time calibration read (#838).
+
+    A non-empty ``game_id`` is added as the ``gameId`` attribute so flagd
+    targeting rules can vary calibration per game (A/B shadow experiments via
+    flag config alone). ``None``/empty adds nothing, so an un-targeted
+    calibration flag resolves IDENTICALLY to today — a pure context addition
+    with no behavior change unless a targeting rule references ``gameId``.
+    """
+    if game_id:
+        return EvaluationContext(attributes={"gameId": game_id})
+    return EvaluationContext()
+
+
+def read_object_flag(domain: str, flag_key: str, default: dict, game_id: str | None = None) -> dict:
     """
     Read an object-typed flag once at init time, with a hardcoded fallback.
 
@@ -193,6 +207,10 @@ def read_object_flag(domain: str, flag_key: str, default: dict) -> dict:
         domain: OpenFeature domain / flagSetId (e.g. "game")
         flag_key: Object flag key (e.g. "thresholds")
         default: Fallback value returned on any error or missing flag
+        game_id: Optional owning game id (#838); added as the ``gameId`` context
+            attribute so flagd targeting can vary calibration per game. Omitted
+            from the context when ``None``/empty — un-targeted reads are
+            unchanged.
 
     Returns:
         The flag's object value, or ``default`` on failure.
@@ -200,7 +218,7 @@ def read_object_flag(domain: str, flag_key: str, default: dict) -> dict:
     try:
         init_flag_domain(domain)
         client = get_flag_client(domain)
-        value = client.get_object_value(flag_key, default, EvaluationContext())
+        value = client.get_object_value(flag_key, default, _calibration_context(game_id))
         # get_object_value may return the default sentinel itself; either is fine.
         return value if isinstance(value, dict) else default
     except Exception as e:
@@ -238,7 +256,7 @@ def read_object_flag_variant(domain: str, flag_key: str, targeting_key: str, def
         return default
 
 
-def read_float_flag(domain: str, flag_key: str, default: float) -> float:
+def read_float_flag(domain: str, flag_key: str, default: float, game_id: str | None = None) -> float:
     """
     Read a number-typed flag once at init time, with a hardcoded fallback.
 
@@ -255,6 +273,7 @@ def read_float_flag(domain: str, flag_key: str, default: float) -> float:
         domain: OpenFeature domain / flagSetId (e.g. "game")
         flag_key: Number flag key (e.g. "death_grace_period_seconds")
         default: Fallback value returned on any error or missing flag
+        game_id: Optional owning game id (#838); added as ``gameId`` context.
 
     Returns:
         The flag's float value, or ``default`` on failure.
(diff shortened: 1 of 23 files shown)


aepfli:
Integration failure root-caused: NOT this PR's bug — a latent NonStop crash on main (`GameplayStreamConfig` still passed the long-removed `serials` proto field; every NonStop game died ~100ms after start). This PR's honest per-session evaluation exposed it: main's intervention manager evaluated against the dead game vacuously (`get_live_game()=None` skipped the mode matrix but still published), so the revive test passed against a corpse. Fix in #899 (with the regression unit test + the `game_error`-is-never-terminal-success tightening); cherry-picked here (`ef34a9cf`) so CI can go green — the duplicate patch drops on rebase after #899 merges. Locally verified: the revive test passes with fix+#838 together.

aepfli:
Second latent crash surfaced by the game_error tightening and fixed on #899 (cherry-picked here): Tournament's match loop called `__anext__()` directly on the stream-stream call — `InterceptedStreamStreamCall` (tracing interceptors) exposes `read()`/`__aiter__` but no direct `__anext__`, fatal since the grpcio 1.81 bump (#846). Switched to the canonical `read()` + `grpc.aio.EOF`. Same three-mask pattern as NonStop: write-only unit mock, game_error-as-terminal, vacuous intervention evaluation.

PR #899, opened 2026-06-06
test: Swapper end strategy must not mistake an unregistered kill for game end

aepfli:
## Summary

Every NonStop game has been crashing ~100ms after `game_started` on main:

```
Game loop error: Protocol message GameplayStreamConfig has no "serials" field.
```

`nonstop_joust.py`'s dynamic-filter stream init still passed `serials=[]` — a proto field removed and reserved ages ago (PR #58, "use colors instead"). Fixed by mirroring the base class (initial `colors` carry the player set; the alive-set filter goes via `FilterUpdate`).

## How it stayed hidden (three masks)

1. **Unit tests** inject a mock `gameplay_stream` *past* the init block → the real proto construction was never exercised. Now extracted as `_build_initial_stream_control()` with a regression test.
2. **The parallel lifecycle batch** (#826) counted `game_error` as a valid terminal event → a NonStop that crashes instantly "ended successfully". `game_error` now fails the mode explicitly.
3. **The intervention manager** evaluated against the dead game vacuously (`get_live_game()=None` skipped the mode matrix but still published) → `test_revive_allowed_in_respawn_mode` passed against a corpse.

**#838's honest per-session evaluation (PR #890) is what exposed it** — its integration failure was this bug, not #838's. #890 needs this fix on its base to go green.

Verified: 41 nonstop unit tests (1 new regression test), full game_coordinator suite 926 green, lint clean. Root-caused via local repro (`SKIP_TEARDOWN=1 make test-dev TEST=test_revive_allowed_in_respawn_mode`) + coordinator logs.


aepfli:
**PR slimmed to its true remaining delta.** Everything else here (NonStop crash fix, both Tournament crash fixes, the game_error tightening, CI calibration flags, the demotion to #903) reached main inside #890's squash via the cherry-pick train. What remains is the one commit that post-dates the merge: the Swapper strategy treating an unregistered kill (grace/EMA, #757 family) as "game ended" and silently abandoning a live 2v2 — the ~50% flake (failed here while identical code passed on #890 minutes apart; the failure player-map shows zero net team progress, ruling out bounce-back for this failure mode). Re-queries to distinguish, retries within the deadline.

Issue #916, opened 2026-06-10
Agent: rolling game narrative — accumulate per-game history for LLM prompts


aepfli:
## Context
The project pitch says the agent observes "OpenTelemetry spans aggregated into rolling game narratives". What's built is a **live snapshot**: `GameContext` is overwritten per signal; the only history kept is the elimination sequence. LLM prompts (#739) and the post-game retro (#844) therefore see point-in-time state — trends are invisible ("player X has been fading for 90s", "the last tempo bump changed nothing").

## Proposal
Bounded per-game-partition history (ring buffer), accumulating:
- periodic state deltas (active player count, intensity aggregates, fitness evaluations)
- dispatched **and blocked** interventions with timestamps
- session phase transitions (start, eliminations, end)

Surface a compact timeline section in the prompt builder (`llm/prompt.go` User part) and retro prompt (`llm/retro.go`).

## Acceptance
- Prompt contains a bounded recent-events timeline per game
- Memory bounded (fixed capacity per partition, evicted with session)
- Per-game isolation preserved (#845 multiplexer)
- Unit tests for accumulation, capacity, eviction

Refs: #739, #844, #845. From the 2026-06-10 agent-layer gap evaluation.

aepfli:
Core delivered in #934 (merged to main): a bounded (64-entry) per-game ring-buffer narrative — state deltas (deduped), eliminations, and phase transitions — surfaced as a "Recent events" timeline in the LLM prompt and retro builders. All four acceptance criteria met (bounded timeline, bounded memory, per-game isolation, unit tests for accumulation/capacity/eviction); sub-agent review confirmed correct ring math, concurrency, isolation, and determinism.

Two follow-up threads tracked separately:
- #943 — elimination dedupe correctness fix (cross-store double-narration found in review; PR open, ready to merge).
- #945 — wiring dispatched/blocked interventions into the timeline (the `AppendInterventionEvent` seam was built + tested in #934 but the Loop→Store wiring was deferred).

Closing the core issue per the manual-close convention; the threads above carry the remainder.

Issue #845, opened 2026-06-05
Per-game context partitioning: agent decisions keyed by game_id


aepfli:
## What to build

Partition the agent's OBSERVE→DECIDE state by `game_id` so concurrent games (real since the M6 multi-session coordinator, #775/#776) each get their own context, decision evaluation, and intervention budget. Agent-side companion to #838 (which does the same for the coordinator's InterventionManager / flag evaluation context).

## Current behavior (verified)

- `gamecontext.Store` holds exactly **one** context with one `Session` (services/agent/gamecontext/store.go:21-93). Concurrent games collapse into it: players from different games merge into one `Players` map, both games' `game_active`/`duration` signals overwrite each other, and `AdoptSessionID` (store.go:249) is last-writer-wins on whichever `game_id` label arrived most recently.
- The weighted rate limiter (services/agent/decision/ratelimit.go) is **one global instance** — `policy.max_interventions_per_minute = 2` would be shared across all concurrent games.
- `OnEvaluate` runs per OTLP Export with the single merged context; `decision.throttle_seconds` only throttles logs/capture spans, not decisions.

So with multiple games the agent doesn't decide per-game — it sees one blurred game and spends one budget.

## Scope

- Key context stores by `game_id` (labels exist on exported metrics/spans since #776; `game_kind` distinguishes real vs shadow): extraction routes signals into the right per-game store
- Per-game loop state: rate-limiter budget, throttle, LayerState, session lifecycle/eviction (reuse `lifecycle.*` flags semantics per game)
- `gate.ShouldEvaluate` per game; an ended game's partition evicts after grace
- `game_id` (and `game_kind`) on every decision span and in the #739 LLM prompt's game snapshot
- Signals that arrive without a `game_id` label: define and document the fallback (single legacy partition)
- Per-player interventions already target serials; serial→game membership comes from the partitioned roster

## Acceptance Criteria

- [ ] Two concurrent games produce two independent GameContexts — no cross-game player or session-signal bleed
- [ ] Each game gets its own `max_interventions_per_minute` budget
- [ ] Decision spans and LLM prompt snapshots carry `game_id` + `game_kind`
- [ ] A game ending evicts only its own partition; the other game's decisions continue unaffected
- [ ] Signals without `game_id` fall back deterministically (documented)
- [ ] Shadow game running alongside a real game: agent decisions for each are independently attributable in Jaeger

[...]

Issue #844, opened 2026-06-05
Post-game analysis: LLM retrospective suggests calibration tweaks


aepfli:
## What to build

A **post-game analyst** path: when a game ends, the agent builds a retrospective prompt from the full session and asks the LLM to suggest **calibration tweaks for the next game**. This is the LLM use case that tolerates cloud latency — in-game decisions need fast local inference, but between games even a slow backend (or `claude -p` under a Max subscription) works.

## Why

Today the agent stops entirely at game end: `gate.ShouldEvaluate` (services/agent/gate/gate.go:22) requires `GameActive == true` plus a fresh active player. Yet game end is exactly when the richest data exists (full elimination sequence, final duration, skill spread) and when suggestions can take effect — the #766 calibration flags (threshold tables, `global_difficulty_factor`, `pacing_profile`) are read at game **init**.

## Scope

- Trigger on the `GameActive` true→false transition (the hook exists: `Store.SetGameActive`, services/agent/gamecontext/store.go:222; session state is retained through `sessionGrace`)
- Retrospective prompt in `services/agent/llm` (sibling of the in-game builder from #739): full-session summary + the calibration surface as the suggestion vocabulary + JSON response contract for suggested tweaks
- **Capture-first, same pattern as the #739 spike**: until #741/#742 land, emit the prompt on a dedicated `agent.llm.retro` span (replayable via `services/agent/scripts/replay-prompt.sh`); wire a real backend once the fallback chain exists
- Suggestions are recorded, **never auto-applied** in this issue — applying via `FlagConfigWriter` behind a flag is a follow-up
- Post-game path must not run during an active game and must not consume the in-game intervention budget

## Acceptance Criteria

- [ ] Game end (GameActive true→false) triggers exactly one retrospective per session
- [ ] Retrospective prompt contains the full elimination sequence, duration, and per-player end-state
- [ ] Prompt captured on an `agent.llm.retro` span with the same attribution attributes as `agent.llm.prompt` (inference.configured/used/fallback_reason)
- [ ] Suggested-tweaks response contract maps to the calibration flag surface (#766); contract documented
- [ ] No retrospective during active games; no interaction with the per-minute intervention budget

Part of M4. Related: #739 (prompt builder reused), #741 (backend resolution), #766 (calibration flags).

Issue #739, opened 2026-06-04
LLM decision path with objective-aware prompting


aepfli:
## What to build

The LLM decision path: `llm_decide(context, backend, objectives)` — the `agent.mode = "llm"` branch of the decision loop, with **objective-aware prompting**.

## Scope

- `llm_decide` takes the extracted game context, the resolved inference backend, and the objective weights from `agent.objectives`
- Prompting is objective-aware: the weighted objectives (`endurance`, `balanced`, `accelerate`, `chaos`) shape what the model is asked to optimize for
- LLM output maps to the same action structure as the rules engine (`type`, `reason`, `objective_served`, `fitness`) so span attribution and the permission gate work identically
- The permission layer still applies: LLM-chosen actions not in `interventions.allowed` are blocked with `decision.blocked = true`
- Decision spans carry `inference.configured` and `inference.used`

## Acceptance Criteria

- [ ] `agent.mode = "llm"` routes decisions through the LLM path
- [ ] Prompts incorporate the current objective weights and game context
- [ ] LLM responses parse into valid actions; unparseable/invalid output does not dispatch arbitrary actions
- [ ] Permission gating and rate limiting apply identically to LLM decisions
- [ ] Decision spans show the LLM backend used and the reasoning behind the action

aepfli:
Delivered in #946 (merged to main): the `agent.mode="llm"` decision path — `llmDecide` calls the resolved backend, strictly decodes the response into a `Decision`, and runs it through the same permission/rate-limit chain as rules; unparseable/invalid output never dispatches an arbitrary action (falls back to rules). Reviewed clean. Closing per the manual-close convention.

PR #934, opened 2026-06-12
feat(agent): bounded per-game rolling narrative for LLM prompts (#916)

aepfli:
## Summary

Part of #916. The agent kept only a **live snapshot** per game — `GameContext` is overwritten per signal, with the elimination sequence the sole retained history — so the in-game prompt (#739) and the post-game retro (#844) saw point-in-time state with no trend. This adds a **bounded per-game-partition rolling narrative** (ring buffer) and surfaces it in both prompt builders.

## Scope

- **`gamecontext/timeline.go`** — a FIXED-capacity (64) ring buffer of timestamped, typed events (`state` / `elimination` / `phase` / `intervention`). It lives ON the per-game `Store`, guarded by the existing Store mutex, so it is automatically **per-partition (#845)** and evicted with the partition. Append past cap evicts the oldest; the snapshot copy shares no backing array with the live ring.
- **`Store`** appends from what it already observes: periodic **state deltas** (active player count, mean movement intensity — deduped so a steady game does not churn the ring), **eliminations** (with 1-based order), and **session phase transitions** (start/end). The narrative is cleared on session restart and on grace eviction, matching `EliminationSequence`'s session scope.
- **`GameContext.Timeline`** carries a bounded, oldest-first snapshot copy, so `llm.Build` / `llm.BuildRetro` stay pure and deterministic (no signature change beyond the field on the context they already receive).
- **`llm/prompt.go` + `llm/retro.go`** render a compact `Recent events` section — one line per event, relative ages — via `gamecontext.RenderTimeline`. Empty timelines render `(none)`.

## Acceptance

1. ✅ Prompt contains a bounded recent-events timeline per game (golden tests `timeline_3players` / `retro_timeline`).
2. ✅ Memory bounded — fixed `timelineCap` (64) ring buffer, oldest evicted at cap.
3. ✅ Per-game isolation preserved — history lives on the per-game Store (multiplexer isolation test).
4. ✅ Unit tests for accumulation, capacity, eviction (session reset + grace + partition removal), isolation, and deterministic rendering.

## Deferred (with rationale)

**Intervention-event capture into the timeline.** The decision `Loop` — not the `Store` — knows the dispatched-vs-blocked outcome, and the two are separate objects. `Store.AppendInterventionEvent(intervention, serial, blocked)` is the clean seam (the receiver pipeline holds both the partition Store and the Loop), and it is implemented + unit-tested here. Wiring it into the per-decision action path is left for a follow-up referencing #916 to keep this PR a tight, well-tested first cut over what the Store already sees. The `intervention` event kind and renderer are already in place, so the follow-up is pure wiring.

## Tests

`gofmt -l` clean, `go vet ./...` clean, `go build ./...` clean, `go test -race ./...` all green from `services/agent`.

### DIFF
--- a/services/agent/gamecontext/model.go
+++ b/services/agent/gamecontext/model.go
@@ -90,4 +90,13 @@ type GameContext struct {
 	Session   SessionSignals
 	Players   map[string]*PlayerSignals
 	UpdatedAt time.Time
+
+	// Timeline is a bounded, oldest-first slice of the partition's recent rolling
+	// narrative (#916): periodic state deltas, eliminations, and session phase
+	// transitions, each timestamped. It is a SNAPSHOT copy of the Store's ring
+	// buffer taken in Snapshot(), so it shares no backing array with the live ring
+	// and is safe to render after later Store mutations. nil when nothing has been
+	// recorded yet. The llm package renders it into a compact timeline section in
+	// both the in-game prompt (#739) and the retro (#844) via RenderTimeline.
+	Timeline []TimelineEvent
 }
--- a/services/agent/gamecontext/multiplexer_test.go
+++ b/services/agent/gamecontext/multiplexer_test.go
@@ -350,3 +350,55 @@ func TestMux_SpanRouting(t *testing.T) {
 		t.Error("P2 bled into game-A")
 	}
 }
+
+// TestMux_TimelineIsolation verifies the #916 rolling narrative is per-partition:
+// two concurrent games accumulate independent timelines, and one game's
+// eliminations/state deltas never appear in the other's. This is the per-game
+// isolation acceptance criterion — the timeline lives on the per-game Store, so
+// it inherits the multiplexer's partition isolation for free.
+func TestMux_TimelineIsolation(t *testing.T) {
+	clk := &muxClock{t: time.Unix(1000, 0)}
+	mux := newTestMux(clk, time.Hour, time.Hour, nil, nil)
+
+	// Game A: start + two distinct intensities (state deltas) + an elimination.
+	mux.ApplyMetrics(gameActiveFor("game-A", 1))
+	mux.ApplyMetrics(playerIntensityFor("PA", "game-A", 1.0))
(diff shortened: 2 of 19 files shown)


aepfli:
Reviewed — clean ring buffer, deterministic rendering, good per-partition isolation and dedupe of state deltas. One latent non-blocker: `SetEliminationOrder` dedupes elimination *narration* against the `elimOrder` map while `appendElimination` (deaths/span paths) dedupes against `EliminationSequence`. If a deployment ever emits **both** a death/span elimination signal and a preferred order signal for the same serial, and the death/span path lands first, `SetEliminationOrder`'s `!alreadyKnown` check passes and the serial gets a **second** timeline line (the existing `EliminationSequence` rebuild also wouldn't suppress it). Today only one source is active per deployment so it's latent — but when the preferred `game_player_elimination_order` source lands alongside the fallback, consider unifying narration dedupe on a single set. Observability-only, not a merge blocker.

aepfli:
Sub-agent review complete: **approve with suggestions**, no strict blockers — `go test -race ./...`, `gofmt`, `go vet` all green; ring math, snapshot shared-nothing guarantees, per-game isolation, dual reset (restart + grace), and determinism all verified correct.

The one real defect was the elimination double-narration (confirmed empirically reproducible: 2 lines for one serial). Per the reviewer's recommendation to fix it now rather than defer — the trap is easy to forget once `game_player_elimination_order` ships — I pushed a fix (commit on this branch): `SetEliminationOrder` now guards narration on `EliminationSequence` membership in addition to the `elimOrder` map, so the two source paths can't each narrate the same serial. Added `TestStore_TimelineEliminationNoDoubleNarration` covering both report orderings (span-then-order and order-then-span). Full suite green.

Remaining review nits are cosmetic (golden fixture elimination-order vs sequence ordering; unreachable `renderEvent` default) and left as-is. The deferred intervention-event timeline wiring (the `AppendInterventionEvent` seam) remains a noted follow-up.

Issue #847, opened 2026-06-05
LLM call gating: eligibility, cadence, and global budget


aepfli:
## What to build

A three-layer, flag-controlled gate in front of every LLM decision call, so the llm path (#739) can never burn tokens uncontrolled. **Prerequisite of #739's `llm_decide` — no LLM request without passing this gate.**

## Why

`OnEvaluate` runs on every OTLP Export (no decision throttle exists — `decision.throttle_seconds` only gates logs/capture spans). With per-game loops (#845) and 4 concurrent games (e.g. idle shadow games, whose mock controllers look fully active to `gate.ShouldEvaluate`), a naive llm path would fire several LLM requests **per second**: instant Claude Max rate-limit lockout, and more requests than Jetson-hosted phi4-mini/gemma3:4b can serve (1–5s per inference).

## The three layers (all flagd, hot-tunable, capability/permission domain)

| Layer | Flag | Default | fallback_reason when gated |
|---|---|---|---|
| Eligibility — which games may use LLM | `llm.eligible_game_kinds` | `["real"]` (shadow games rules-only) | `llm_not_eligible` |
| Cadence — per game | `llm.min_decision_interval_seconds` | `10` (≤1 LLM call per game per interval) | `llm_interval` |
| Global budget — across all games | `llm.max_requests_per_minute` | `6` (shared cap, mirrors the weighted intervention budget in `decision/ratelimit.go`) | `llm_budget_exhausted` |

Worst case with defaults: 6 LLM calls/min total regardless of game count; 4 idle shadow games cost zero tokens.

## Scope

- Gate evaluated before any backend call; a gated cycle falls back to the rules engine with full span attribution (`inference.configured=<model>`, `inference.used="rules"`, `inference.fallback_reason=<reason above>`) — every skipped LLM call is visible in Jaeger
- Per-game cadence + eligibility state lives with the per-game loop state (#845); the global budget is one shared limiter (concurrency-safe, same pattern as the unified intervention rate limiter)
- The post-game analyst (#844) is once-per-session by construction but must draw from the same global budget (a retrospective is one LLM call)
- Future refinement (out of scope, note in docs): significance-triggered escalation — rules engine as cheap filter, LLM consulted only on noteworthy events (elimination, fitness violation, variance spike)

## Acceptance Criteria

- [ ] No code path can issue an LLM request without passing eligibility → cadence → budget, in that order
- [ ] Shadow games produce zero LLM calls with default flags
[...]

Issue #740, opened 2026-06-04
Prompt variants per flag value


aepfli:
## What to build

Prompt variants selected per flag value — the capability layer's `agent.prompt_variant` flag controlling how the LLM is prompted:

```
agent.prompt_variant = "conservative" | "aggressive" | "balanced"
```

## Scope

- Three prompt variants for the LLM decision path, selected at decision time from the flag
- Variant changes are live: flipping `agent.prompt_variant` on stage changes agent behavior on the next decision, no restart
- The variant in effect is part of the decision's audit trail (capability-layer flag state on the decision span)

## Acceptance Criteria

- [ ] All three variants (`conservative`, `aggressive`, `balanced`) implemented for the LLM path
- [ ] Variant selection read from the flag on each decision
- [ ] Flipping the flag at runtime observably changes decision behavior
- [ ] The active variant is recorded on the decision span

aepfli:
Delivered in #947 (merged to main): the three `agent.prompt_variant` values (conservative/aggressive/balanced) produce materially different System prompts on the LLM path, selected live per-decision, with the effective variant on the decision span and a safe conservative fallback for unknown values. Reviewed clean. Closing per the manual-close convention.

Issue #741, opened 2026-06-04
Inference availability checking and fallback chain


aepfli:
## What to build

Inference availability checking and the fallback chain — the agent uses whatever intelligence is available and degrades gracefully to rules:

```
configured model (claude | copilot)
    ↓ unreachable
gemma3:4b on Jetson
    ↓ unreachable
phi4-mini on localhost
    ↓ unreachable
rules engine
```

## Scope

- `resolve_backend()`: walk the chain from the configured model (`agent.model` flag) down to the rules engine
- Availability checked **periodically, cached between checks** — not probed on every decision
- Every decision span carries the inference attribution triple:
  - `inference.configured` — what the flag asked for (e.g. `"claude"`)
  - `inference.used` — what actually decided (e.g. `"rules"`)
  - `inference.fallback_reason` — why (e.g. `"endpoint_unreachable"`)
- Chain bottoms out at the rules engine, so the system always decides — fully standalone with no Jetson and no cloud

## Acceptance Criteria

- [ ] Backend resolution follows the documented chain order from the configured model
- [ ] Availability checks run periodically with results cached between checks
- [ ] All three inference attributes present on every decision span, including when configured == used
- [ ] Unplugging the Jetson mid-session degrades to localhost phi4-mini, then to rules — without errors or stalled decisions
- [ ] Recovery: when a higher tier becomes reachable again, the agent moves back up the chain

aepfli:
Delivered in #942 (merged to main): the inference availability resolver + fallback chain (cloud→Jetson→localhost→rules) with cached periodic probing and the honest inference.configured/used/fallback_reason triple on every decision span. Reviewed clean. Closing per the manual-close convention.

PR #938, opened 2026-06-12
feat(agent): LLM call gating — eligibility, cadence, global budget (#847)

aepfli:
Part of #847. Adds a three-layer, flag-controlled gate in FRONT of every LLM decision attempt (#739), so the future llm path can never burn tokens uncontrolled. Prerequisite of #739.

## Scope

A gate evaluated EVERY cycle (never cached), in this exact order, all flagd-driven and hot-reloaded:

| Layer | Flag (`agent.json`) | Default | fallback_reason when gated |
|---|---|---|---|
| Eligibility | `llm.eligible_game_kinds` | `["real"]` | `llm_not_eligible` |
| Cadence (per game) | `llm.min_decision_interval_seconds` | `10` | `llm_interval` |
| Global budget | `llm.max_requests_per_minute` | `6` | `llm_budget_exhausted` |

- **Gate DENIED**: the cycle does NOT build/capture the prompt (no token burn, no serialization), records the specific reason on the decision span's `inference.fallback_reason`, and increments `agent_llm_gated_total{reason=...}`. The rules engine still runs (the gate only guards the LLM *attempt*).
- **Gate ALLOWED**: unchanged behavior — capture prompt -> rules fallback with `no_backend_available`.
- The per-cycle resolved fallback reason now flows onto the decision span via a new `LayerState.LLMFallbackReason` field (replacing the static `inferenceAttribution` for mode=`llm`); mode=`rules` is unchanged.

### Wiring
- New `llmBudget` (global, sliding 1-min window, mutex-guarded, unit-cost-per-request) mirrors `decision/ratelimit.go`. Constructed ONCE in `main.go` and injected into every per-game `Loop` (and the retro coordinator) via `SetLLMBudget`, so all concurrent games share one cap.
- Per-game cadence (`lastLLMAttempt`) lives on the `Loop`.
- New `flags.LLMGate` struct populated in `Flags.Evaluate`; numeric flags use the typed getters (float seconds -> Duration, int budget); eligibility array uses `ObjectValue`/`toStringSlice`, fail-closed to `["real"]`.
- New `agent_llm_gated_total` counter (no-op default so existing Loop tests need no wiring).
- **#844 retrospective** is one LLM request and now draws from the SAME shared global budget; an exhausted budget at game end gates the retro (no span).

## Acceptance criteria

- [x] No LLM attempt without passing eligibility -> cadence -> budget, in order
- [x] Shadow games produce zero LLM attempts with default flags (`TestGate_ShadowGameNotEligible`)
- [x] 4 concurrent eligible games never exceed `max_requests_per_minute` combined (`TestGate_SharedBudgetAcrossLoops`)
- [x] Every gated decision carries the correct `inference.fallback_reason` (`TestGate_*`)
- [x] Flags hot-reload: tightening the budget mid-session takes effect next decision (`TestGate_BudgetHotReload`)
- [x] Gate state observable: `agent_llm_gated_total{reason=...}` increments per gated cycle
- [x] Retro draws from the same global budget (`TestRetroCapture_DrawsFromGlobalBudget`)

## Out of scope (per #847, not built)
Significance-triggered escalation; the real LLM backend call (#739); prompt variants (#740).

## Verification
`gofmt -l .` clean, `go vet ./...`, `go build ./...`, `go test -race ./...` all green from `services/agent/`.

### DIFF
--- a/services/agent/decision/attribution_test.go
+++ b/services/agent/decision/attribution_test.go
@@ -25,6 +25,10 @@ func recordingLoop(t *testing.T, snap flags.Snapshot, out []Decision) (*Loop, *t
 	l.Tracer = tp.Tracer("test")
 	l.Rules = fakeRules{out: out}
 	l.Actions = &fakeSink{}
+	// Inject a fresh global LLM budget (#847) so the gate's budget layer can admit
+	// an llm-mode cycle; without it a nil budget would gate every llm attempt and
+	// the capture-path tests (which assert the ALLOWED behavior) would never fire.
+	l.SetLLMBudget(NewLLMBudget())
 	return l, sr
 }
 
@@ -192,6 +196,14 @@ func TestInferenceAttribution_LLMFallback(t *testing.T) {
 				Capability:           flags.Capability{Model: "phi4-mini"},
 				InterventionsAllowed: []string{"play_audio_cue"},
 				Policy:               flags.Policy{MaxInterventionsPerMinute: 10},
+				// Gate-ALLOWED config (#847): the llm case asserts the unchanged
+				// no_backend_available fallback, which only holds when the gate admits —
+				// so this game kind ("") is eligible, no cadence floor, ample budget.
+				LLMGate: flags.LLMGate{
+					EligibleGameKinds:    []string{""},
+					MinDecisionInterval:  0,
+					MaxRequestsPerMinute: 100,
+				},
 			}
 			l, sr := recordingLoop(t, snap, []Decision{{Intervention: "play_audio_cue"}})
 
--- a/services/agent/decision/decision.go
+++ b/services/agent/decision/decision.go
@@ -44,6 +44,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	otelmetric "go.opentelemetry.io/otel/metric"
 	semconv "go.opentelemetry.io/otel/semconv/v1.34.0"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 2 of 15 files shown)


aepfli:
Post-merge sub-agent review (for the record): **no blocking issues** — `go test -race ./...`, `gofmt -l .`, `go vet ./...` all green; concurrency model sound (`lastLLMAttempt` only ever touched under `llmGate`; no lock-order inversion between `llmGate` and the budget mutex), charge-on-admission discipline correct (ineligible/cadence-denied cycles never draw the budget), fail-closed everywhere (nil budget / cap≤0 / empty eligibility), flags read live each cycle, and the retro shares the one global budget. Tests assert real behavior (injected clock, real goroutines).

Five cosmetic nits, none blocking:
1. `decision.go:757` — the `fallback=="" → FallbackNoBackend` default in `layerStateAttributes` is unreachable from the live loop (`decide` always assigns a definite reason).
2. `decision.go:798` — `inferenceAttribution(mode)` is now only called for `mode!="llm"`; its `if mode=="llm"` branch is dead for live callers (retained for the table test).
3. `gateLLM` doc could state that budget is charged on admission **regardless of the `emit` throttle** (correct behavior; saves the next reader a double-take).
4. No explicit test that game A hitting its cadence floor doesn't perturb game B's `lastLLMAttempt` (design trivially guarantees it; per-Loop state).
5. `flags.go:233` — "symmetrical with InterventionsAllowed" applies to the empty-list semantics, not the parse-failure default (`["real"]` vs `nil`).

Nits 1–3 land in the attribution code that #741 reworks — they'll be cleaned/verified there. Filing here only as a record since this is already merged.

Issue #917, opened 2026-06-10
Agent: async LLM inference — latency budget and stale-decision application


aepfli:
## Context
The decision loop runs ~1 Hz with per-cycle flag evaluation; an LLM call takes 2–10s. There is no pattern yet for handling a decision that arrives seconds after the context that produced it.

## Needed
- Fire inference async without blocking the per-game loop
- On completion, **re-validate** against current context before applying: game still active, target player still alive/connected, permission + rate budget still available
- Discard-with-reason path (span attribute, e.g. `inference.discarded_reason=stale_context`)
- Define the latency budget after which a result is dropped outright

Distinct from but coordinating with #847 (eligibility/cadence/global budget decide *whether/when to call*; this issue is *what to do with the answer*). Also feeds #741 (fallback chain: timeout → rules).

## Acceptance
- Loop never blocks on inference
- Stale results provably discarded (test with game-end during inference)
- Audit span records inference latency + applied/discarded outcome

Refs: #739, #741, #847. From the 2026-06-10 agent-layer gap evaluation.

aepfli:
Delivered in #949 (merged to main): async LLM inference — `fireInferAsync` so the per-game loop never blocks, re-validation against the CURRENT context at apply time with typed discard reasons (stale_context / game_ended / target_gone / permission_revoked / rate_limited / timeout), the `llm.latency_budget_seconds` flag, and the `agent.llm.apply` span carrying inference latency + applied/discarded outcome. Adversarially reviewed (no data races, in-flight guard airtight, no goroutine leak on shutdown). All ACs met. Closing per the manual-close convention.

Issue #742, opened 2026-06-04
Cloud backends: Claude and Copilot as optional layer


aepfli:
## What to build

Cloud inference backends — **Claude and Copilot** — as an optional top layer of the inference chain, selectable via the capability flag:

```
agent.model = "phi4-mini" | "gemma3:4b" | "claude" | "copilot"
```

## Scope

- Claude and Copilot integrations for `llm_decide`, used when configured via `agent.model` and reachable
- Strictly optional: these sit at the top of the fallback chain; when unreachable, the chain falls through to gemma3:4b on the Jetson → phi4-mini on localhost → rules engine
- The core system keeps its no-cloud-dependency property — cloud backends are an optional layer, never a requirement
- Same action contract and span attribution as the local LLM path (`inference.configured = "claude"`, `inference.used`, `inference.fallback_reason`)

## Acceptance Criteria

- [ ] `agent.model = "claude"` and `agent.model = "copilot"` route decisions to the respective cloud backend when reachable
- [ ] Cloud unreachability triggers clean fallback down the chain with `inference.fallback_reason` set
- [ ] System remains fully functional with no network/cloud access
- [ ] Cloud-backed decisions carry the same span attribution as local ones

aepfli:
## Recommendation: Use API Key + Sidecar Proxy (Option A)

After researching Claude Max subscription programmatic options, **Claude Max OAuth tokens cannot be used for automated inference backends due to Anthropic's 2026 ToS restrictions**. The recommendation is:

### ✅ Recommended Path: Anthropic API Key + HTTP Sidecar Proxy

**Why this is the right choice:**

1. **Removes the OAuth ToS violation entirely** — API keys are the sanctioned programmatic path; no ambiguity.
2. **Decouples auth from the agent container** — The agent (Go binary, alpine, no CLI) stays clean. Auth lives on the host where the Max subscription's OAuth already works.
3. **Minimal wiring** — A small sidecar or the Claude Code CLI itself (if running locally) exposes an HTTP/gRPC endpoint that the agent's `Backend.Infer` calls.
4. **Cost-effective for steady proposals** — Each proposal runs ~1–3 times per minute (gated by `code_improvement.min_interval_seconds`). Estimated ~100–150 proposals/day = negligible API cost, well within Max subscription credit (`$100/month` for Max 5x = ~$3.33/day freely available).

### Concrete Wiring (Option A – sidecar approach)

\`\`\`go
[...]

Issue #738, opened 2026-06-04
Ollama on Jetson: phi4-mini and gemma3:4b


aepfli:
## What to build

Ollama inference serving on the Jetson Orin Nano Super, hosting **phi4-mini** and **gemma3:4b** for the agent's LLM decision path.

## Hardware context

- Jetson Orin Nano Super Developer Kit with 256GB NVMe SSD
- Agent runs on the Pi 5 and calls the Jetson over the local network for inference
- The Jetson is **optional** — the system runs fully standalone without it (fallback chain handles unavailability)
- All hardware on the demo table, local network only, no cloud dependencies

## Scope

- Ollama installed and serving on the Jetson, models stored on the NVMe
- `gemma3:4b` and `phi4-mini` pulled and runnable
- Reachable from the Pi 5 over the local network (Cat6 direct/switch)
- Note: per the fallback chain, `phi4-mini` also needs to run on localhost (Pi 5) as the second fallback tier — this issue covers provisioning both serving locations

## Acceptance Criteria

- [ ] Ollama serving on the Jetson with `gemma3:4b` and `phi4-mini` available
- [ ] Agent on the Pi 5 can complete an inference request against the Jetson over the local network
- [ ] `phi4-mini` runnable on localhost (Pi 5) as the next fallback tier
- [ ] Unplugging the Jetson does not break the system (degrades per the fallback chain)

aepfli:
Superseded by the ratified #1046 design. The Ollama backend is now GENERIC: #1048 implements an OpenAI-compatible Backend.Infer pointed at any Ollama (dev host or a Jetson over LAN) via AGENT_INFERENCE_BASE_URL/_MODEL; #1049 adds the gateway for fan-out. A Jetson is just a base_url + deployment target — a Jetson-specific hardware deployment, if/when needed, can be a fresh deployment issue. Closing as subsumed by #1048/#1049.

Issue #924, opened 2026-06-10
Agent: interventions.json corruption recovery — poisoned-file detection, self-heal, alert


aepfli:
## Context
`actions/writer.go` does in-place read-modify-write on `/etc/flagd/interventions.json` (deliberately no temp+rename — avoids EBUSY on docker bind mounts). A crash mid-write can leave invalid JSON; flagd then rejects the file and **all interventions die silently** — every flag read falls back to defaults.

The existing alert (`agent_intervention_writes_total{result="error"}`) catches write *errors*, not a previously-poisoned file that makes flagd serve stale/default state.

## Proposal
- Writer: validate-after-write (re-parse; on failure restore last-known-good from an in-memory copy)
- Startup: validate the file; self-heal to a neutral document if corrupt
- Alert on flagd sync/parse failures (flagd logs/metrics) so a poisoned file is visible within seconds

## Acceptance
- Chaos test: truncate the file mid-session → agent restores it and an alert fires
- Documented recovery runbook entry

From the 2026-06-10 agent-layer gap evaluation. Refs: #803 (self-observability), #831 (runbook).

PR #942, opened 2026-06-12
feat(agent): inference availability checking and fallback chain (#741)

aepfli:
Part of #741. Builds the inference availability checker + fallback chain: `resolve_backend()` walks from the configured `agent.model` flag down to the always-available rules engine, so the agent uses whatever intelligence is reachable and degrades gracefully:

```
configured model (claude | copilot)   [agent.model flag]
    ↓ unreachable
gemma3:4b on Jetson                    (#738)
    ↓ unreachable
phi4-mini on localhost                 (#739)
    ↓ unreachable
rules engine                           ← always available; the system ALWAYS decides
```

> **Stacking note:** authored as a stacked diff on top of #847 / PR #938 (`feat/847-llm-call-gating`). #938 squash-merged to main while this was in flight, so this branch was rebased onto main; it now targets `main` directly and builds ON #847's merged gate. No #847 changes appear in this diff.

This is the ABSTRACTION + resolver + attribution + periodic-cached probing. There is **no real backend call yet** — #739 implements `llm_decide`, Jetson #738 and cloud #742 are hardware/credential-blocked. So a "reachable" tier means "this is who #739 WILL call"; the rules engine still produces the Decision until then. The resolver's job is to make `inference.used` / `inference.fallback_reason` HONEST on every decision span.

## What's in it

- **Backend abstraction** (`decision/resolver.go`): a `Backend` interface (`Name()` + `Available(ctx) bool`) per tier, an ordered chain, and a `Resolver` holding a periodically-refreshed availability **cache**. `resolve()` reads the cache only — **no probe on the decision hot path** (acceptance #2). A background ticker (`Start`) re-probes every `DefaultProbeInterval` (15s const), so an unplugged tier degrades and a recovered tier is climbed back to within one interval (acceptance #4/#5). Mutex-guarded for the concurrent Export handlers.
- **Production probe** (`decision/probe_backend.go`): a bounded TCP dial per tier; empty endpoint = permanently unreachable. Injectable via the `Backend` interface so tests use fakes (never the network). #739 swaps the dial for a real model-health check.
- **Honest attribution**: `inference.used` / `inference.fallback_reason` are now the resolved tier and degradation reason, threaded through `LayerState.InferenceUsed` (default `rules`) so `layerStateAttributes` emits the resolved value instead of the static `inferenceAttribution(mode)`. `inference.configured` is the model flag. All three present on every decision span, including when configured == used (fallback_reason `""` then).
- **Gate coordination with #847**: `decide()` runs the gate FIRST; `resolve_backend` runs ONLY when the gate ADMITS. A gate-DENIED cycle keeps #847's behavior **exactly** — `inference.used="rules"`, `inference.fallback_reason`=the gate reason — and the resolver never overrides it or consumes anything. New `FallbackEndpointUnreachable` (`endpoint_unreachable`) reason for "a lower tier serves"; `FallbackNoBackend` reconciled to mean "whole chain unreachable, rules decided".
- **Wiring** (`main.go`): ONE shared `Resolver` (one availability cache for the whole agent, like the global budget) injected into every per-game Loop via `SetResolver`; endpoints read from env (`AGENT_CLOUD_ENDPOINT` / `AGENT_JETSON_ENDPOINT` / `AGENT_LOCALHOST_ENDPOINT`) with documented defaults. In dev every endpoint is unreachable, so the chain degrades to rules — exactly the standalone no-Jetson/no-cloud behavior.

## Acceptance criteria

- [x] Backend resolution follows the documented chain order from the configured model (`TestResolve_ChainOrderFromConfiguredModel`)
- [x] Availability checks run periodically with results cached between checks — probe NOT called per decision (`TestResolve_CachedNotProbedPerDecision`, counting fake)
- [x] All three inference attributes present on every decision span, including configured == used (`TestAttribution_ConfiguredEqualsUsed`)
- [x] Unplugging a tier mid-session degrades to the next, then to rules — no errors, no stalled decisions (`TestResolve_DegradesWhenTierUnplugged`, `TestAttribution_DegradesMidSession`)
- [x] Recovery: a recovered higher tier is climbed back to (`TestResolve_RecoversToHigherTier`, `TestAttribution_RecoversMidSession`)
- [x] Gate coordination: a gate-DENIED cycle keeps the gate reason and used=rules; the resolver does not override it (`TestAttribution_GateDeniedKeepsGateReason`, `TestAttribution_GateBudgetExhaustedKeepsGateReason`)

## Out of scope (per #741)
The actual LLM call / `llm_decide` (#739 — stacks on this next). Real Jetson/cloud backends (#738/#742). Async inference (#917). Prompt variants (#740).

## Verification
`gofmt -l .` clean, `go vet ./...`, `go build ./...`, `go test -race ./...` all green from `services/agent/` (9 packages). `make lint` passes.

### DIFF
--- a/services/agent/decision/attribution_test.go
+++ b/services/agent/decision/attribution_test.go
@@ -175,6 +175,43 @@ func TestDisabledSpan_Throttled(t *testing.T) {
 	}
 }
 
+// TestDisabledSpan_LLMModeNoFallbackReason: when the kill switch is off
+// (enabled=false) in llm mode, the disabled span reports inference.used=rules and
+// inference.fallback_reason="" — NOT a fallback reason. The disabled path
+// short-circuits before decide()/resolve_backend run, so no inference was
+// attempted and there is nothing to fall back FROM (#741). This locks the #741
+// attribution change: pre-#741 the static inferenceAttribution(mode) stamped
+// no_backend_available here even though the agent was off; now the resolved
+// LayerState (decide never ran) honestly carries an empty fallback reason.
+func TestDisabledSpan_LLMModeNoFallbackReason(t *testing.T) {
+	snap := flags.Snapshot{
+		Enabled:    false,
+		Mode:       "llm",
+		Capability: flags.Capability{Model: "claude", PromptVariant: "balanced"},
+	}
+	l, sr := recordingLoop(t, snap, []Decision{{Intervention: "play_audio_cue"}})
+
+	l.OnEvaluate(context.Background(), gamecontext.GameContext{SessionID: "s1"}, testTrigger())
+
+	disabled := spansByName(sr.Ended(), SpanDisabled)
+	if len(disabled) != 1 {
+		t.Fatalf("disabled spans = %d, want 1", len(disabled))
+	}
+	d := disabled[0]
+	if v, ok := attrValue(d, AttrMode); !ok || v.AsString() != "llm" {
+		t.Errorf("agent.mode = %q, want llm", v.AsString())
+	}
+	if v, ok := attrValue(d, AttrInferenceUsed); !ok || v.AsString() != InferenceRules {
+		t.Errorf("inference.used = %q, want %q (nothing decided; rules floor)", v.AsString(), InferenceRules)
+	}
+	if v, ok := attrValue(d, AttrInferenceFallback); !ok || v.AsString() != "" {
+		t.Errorf("inference.fallback_reason = %q, want \"\" (agent off, no inference attempted)", v.AsString())
+	}
+	if v, ok := attrValue(d, AttrInferenceConfigured); !ok || v.AsString() != "claude" {
+		t.Errorf("inference.configured = %q, want claude (capability still recorded)", v.AsString())
+	}
+}
+
 // TestInferenceAttribution_LLMFallback: mode=llm records the fallback to rules
 // with a reason; mode=rules records no fallback. Path-agnostic per #729 §5.
 func TestInferenceAttribution_LLMFallback(t *testing.T) {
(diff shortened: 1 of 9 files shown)


aepfli:
Sub-agent review: **mergeable as-is**, no blocking issues. Verified: gate/resolver ordering doesn't regress #847's attribution (on DENY the resolver never runs and the gate reason + `used=rules` stand; on ADMIT it resolves the tier), the availability cache is concurrency-safe (`-race` clean) and read off the hot path with a background ticker, the probe is bounded + leak-free, the lifecycle goroutine stops on ctx cancel, and the two #938 attribution nits are cleaned up.

Addressed the two non-blocking items in a follow-up commit on this branch:
1. **Untested attribution change** — added `TestDisabledSpan_LLMModeNoFallbackReason` locking in that a kill-switched agent in llm mode reports `inference.used=rules` + `fallback_reason=""` (no inference attempted), which is the honest replacement for the pre-#741 static `no_backend_available`.
2. **Synchronous startup probe (~4s on black-holed endpoints)** — `Start()` now runs the first probe inside the background goroutine, so server readiness is never blocked on an optional Jetson/cloud endpoint. `resolve()` serves the rules floor until the first probe completes (harmless — nothing consults the resolver before the #847 cadence floor, and #739 is unwired).
Plus the cosmetic `mu` comment fix (`available` is a map, not a slice). `gofmt`/`go vet`/`go test -race ./...` all green.

aepfli:
Resolved the merge conflict by merging `main` into the branch. The conflict was in `services/agent/main.go` startup wiring, caused by **#941** (#924 interventions.json corruption recovery) independently editing the same region after this branch was cut — unrelated to the #741 stack. Both blocks are kept: the #924 self-heal runs first, then the #741 resolver construction/`Start()`. Build + `gofmt` + `go vet` + full `go test ./...` all green post-merge. Now `MERGEABLE`.

PR #946, opened 2026-06-12
feat(agent): LLM decision path with objective-aware prompting (#739)

aepfli:
Part of #739. Implements `llm_decide` — the `agent.mode == "llm"` branch of the decision loop with **objective-aware prompting**. After #847 (call gate, merged) + #741 (inference resolver, PR #942), a gate-ADMITTED cycle resolved a reachable inference tier but *always ran rules*. This PR makes the agent actually USE that tier: it calls the resolved backend's inference, turns the response into the SAME `Decision` the rules engine produces, and returns it through the existing audit/permission/rate-limit chain.

> **Stacking note:** authored as a stacked diff on top of **#741 / PR #942** (`feat/741-inference-fallback-chain`), which is itself stacked on the merged #847. The base of this PR is `feat/741-inference-fallback-chain`, so the diff shows ONLY the #739 changes on top of #741. If #942 squash-merges to main and its branch is deleted before this merges, rebase onto main (`git rebase --onto origin/main <741-squash-commit> feat/739-llm-decide` — the #741 content is identical and applies cleanly) and retarget at `main`; the diff stays #739-only.

## What's in it

- **Backend gets an inference method.** #741's `Backend` interface (`Name()` + `Available()`) is extended with `Infer(ctx, llm.Prompt) (string, error)`. `Resolution` now carries the resolved `Backend` (nil = rules terminal), so `decide()` branches on a real reachable tier without a second chain walk. `Available()` and `Infer()` stay cohesive: a tier reported reachable is the one we call.
- **`decide()` now calls the tier.** On a gate-ADMITTED cycle where `resolve_backend` returns a reachable non-rules tier (`Backend != nil`), the loop runs `llmDecide` instead of rules. A nil backend (whole chain unreachable — the dev default, no real backend exists) keeps the #741 fallback to rules verbatim.
- **Objective-aware prompt, reused.** `llmDecide` reuses `llm.Build` (the existing builder the capture span uses) — already objective-weighted via the snapshot, so the System prompt encodes the `agent.objectives` weights and the allow-list. The model is asked to optimize for *this* session's objectives (endurance/balanced/accelerate/chaos).
- **Defensive parsing (`llm/decode.go`, new).** A strict response contract `{intervention, target_serial, value, reason, objective_served}`. `Decode` returns a usable `Response` ONLY for well-formed, complete, in-vocab JSON (tolerates surrounding prose / ` ```json ` fences); empty / not-JSON / missing-required-field / out-of-vocab-objective all return typed errors. **Unparseable/invalid/out-of-vocab output NEVER dispatches an arbitrary action** — the cycle falls back to rules with the new `FallbackUnparseable` (`llm_unparseable`) reason, and `inference.used` stays the tier that was called (it answered, just unusably). Exhaustively table-tested.
- **Same gates, identically applied.** The parsed `Decision` flows through the EXACT `runDecision` -> `evaluatePermission` chain as a rules decision. An LLM-chosen action not in `interventions.allowed` is blocked with `decision.blocked=true`; the battery threshold and the weighted rate limit apply unchanged. No LLM-specific dispatch path. A `noop` response is a valid, contract-following choice that dispatches nothing (and is NOT a rules fallback).
- **Spans.** New `agent.llm.infer` span carries the `gen_ai.*` request attribution plus the model's reasoning / objective / action (or `llm.infer.error` on the fallback path). `inference.configured` / `inference.used` (the deciding tier) ride the decision span as before.
- **Production stub.** `endpointBackend.Infer` is a clearly-marked NOT-YET-IMPLEMENTED that returns an error, so a production agent that resolves a reachable endpoint but has no client yet degrades SAFELY to rules. The real Ollama/cloud transport is filled in by #738 (Jetson, hardware-blocked) / #742 (cloud, credential-blocked) — the signature + parse pipeline are already in place.

## Acceptance criteria

- [x] `agent.mode == "llm"` routes through the LLM path when the gate admits AND a backend resolves reachable (`TestLLMDecide_ValidResponseTakesLLMPath`)
- [x] Prompts incorporate the objective weights + game context (`TestLLMDecide_ObjectivesShapePrompt`)
- [x] Unparseable / invalid / out-of-vocab / errored output does NOT dispatch — falls back safely with a recorded reason (`TestLLMDecide_BadResponseFallsBackToRules`, `llm.TestDecode_Rejects`)
- [x] Permission gating + weighted rate limit apply identically (`TestLLMDecide_OutOfAllowListBlocked`, `TestLLMDecide_RateLimitApplies`)
- [x] Decision/infer spans carry `inference.configured`/`inference.used` + the reasoning (`TestLLMDecide_ValidResponseTakesLLMPath`)
- [x] Gate-first ordering preserved: a denied cycle never infers (`TestLLMDecide_GateDeniedNeverInfers`); dev default (unreachable chain) keeps rules (`TestLLMDecide_UnreachableChainKeepsRules`)

## Out of scope (per #739)
Real Jetson/cloud backend clients (#738/#742). Prompt variants per flag (#740 — stacks on this). Async inference (#917 — stacks on this; the call site is structured so it can be wrapped).

## Verification
`gofmt -l .` clean, `go vet ./...`, `go build ./...`, `go test -race ./...` all green from `services/agent/` (9 packages). #847/#741 tests (resolver attribution, gate ordering, disabled+llm) still pass — the #741 `fakeBackend.Infer` now returns a canned valid response so resolver-attribution tests still observe a clean resolved tier.

### DIFF
--- a/services/agent/decision/decision.go
+++ b/services/agent/decision/decision.go
@@ -338,6 +338,13 @@ func (l *Loop) OnEvaluate(ctx context.Context, c gamecontext.GameContext, trig E
 	// after decide so it reflects this cycle's thresholds and live context. This
 	// happens even on an empty-decision cycle so a future cycle-scoped consumer
 	// still sees the evaluation, but it only reaches a span when a decision emits.
+	//
+	// TODO(#739/real-backend): LastFitness is written only inside the rules engine's
+	// Evaluate. On an llm cycle that the backend decided (decide ran llmDecide, not
+	// rules), this reads the PRIOR rules cycle's stale evaluation. Harmless today
+	// (pure-llm sessions never populate it, and there is no real backend), but once a
+	// tier actually decides, a mixed llm/rules sequence would attach stale fitness to
+	// the llm decision span — guard it on "rules ran this cycle" when wiring #917.
 	if rd, ok := l.Rules.(fitnessReader); ok {
 		if vals := rd.LastFitness().Evaluated(); len(vals) > 0 {
 			state.FitnessEvaluated = vals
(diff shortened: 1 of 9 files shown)


aepfli:
Sub-agent review: **no blocking issues** — mergeable. The critical property (untrusted model output can never dispatch an action the rules/permission layer wouldn't allow) is verified in depth: the reviewer actively tried to break it (smuggling a second JSON object → first-`{`/last-`}` span makes it invalid → rules fallback; out-of-vocab/out-of-allow-list intervention → blocked via the SAME `evaluatePermission` path, not a re-impl; malformed/empty/missing-field/infer-error → `FallbackUnparseable` → rules; crafted target/value → only flow downstream after the gates, unknown serial handled like rules). `decide()` not under `l.mu` so a sync `Infer` can't deadlock; `-race` clean. Decoder is linear (no ReDoS), nesting bounded by encoding/json.

Addressed the four non-blocking items in a follow-up commit on this branch:
1. **Real telemetry bug** — `agent.llm.infer` reported the *configured* model, not the tier called; on a degraded chain (`claude`→`phi4-mini`) the request was misattributed. Now uses `backend.Name()`, with a 2-tier degraded test locking it (the single-tier fakes had configured==used so missed it).
2. Refreshed `decide()`'s stale doc header (still described the pre-#739 "no backend yet" spike).
3. `TODO(#738/#742)` at the `Infer` call site to bound it with a per-call timeout once a real transport lands (a slow network `Infer` would otherwise block the Export goroutine — async is #917).
4. `TODO` near the fitness lift for the latent stale-`fitness.evaluated` on a mixed llm/rules sequence (harmless until a real backend; guard when wiring #917).

`gofmt`/`go vet`/`go test -race ./...` all green.

PR #947, opened 2026-06-12
feat(agent): prompt variants per flag value (#740)

aepfli:
Part of #740. Selects HOW the LLM is prompted on the #739 decision path from the capability-layer `agent.prompt_variant` flag:

```
agent.prompt_variant = "conservative" | "aggressive" | "balanced"
```

> **Stacking note:** authored as a stacked diff on top of **#739 / PR #946** (`feat/739-llm-decide`), which itself stacks on the merged #847 + #741. The base of this PR is `feat/739-llm-decide`, so the diff shows ONLY the #740 changes. If #946 squash-merges to `main` and its branch is deleted before this merges, rebase onto main (`git rebase --onto origin/main feat/739-llm-decide feat/740-prompt-variants`) and the PR auto-retargets to `main`; the diff stays #740-only.

## What's in it

- **Three materially-different variants.** The #739 prompt builder already had a `prompt_variant` seam, but the three values were placeholder one-liners that differed only by a recorded label. This makes each variant a distinct multi-line System-prompt block that shifts the agent's RISK POSTURE on top of the existing objective-awareness (objectives + allow-list survive in every variant):
  - `conservative` — HIGH bar to act; bias toward soft/no intervention, ambient cues over state changes.
  - `aggressive` — LOW bar to act; readier to dispatch stronger, state-changing interventions.
  - `balanced` — act on moderately-actionable signals; weigh disruption vs. benefit, no bias either way.
  The differences are in the System prompt TEXT, assertable in tests and reviewed in the goldens.
- **Per-cycle, live.** `llm.Build` resolves the variant on every `OnEvaluate`, so flipping `agent.prompt_variant` on stage changes behavior on the NEXT decision with no restart. Nothing caches the variant. A decision-level test (`TestLLMDecide_VariantReachesInferencePrompt`) flips the flag and asserts the System prompt the real inference call receives changes accordingly.
- **One fail-safe fallback point.** `ResolveVariant` is exported as the SINGLE place the unknown→`conservative` fallback lives, shared by `llm.Build` AND the decision span (`layerstate.go`). A garbage/empty `prompt_variant`:
  - still produces a DEFINED prompt — the inference call is never empty/undefined (`TestLLMDecide_UnknownVariantFallsBackSafely`), and
  - records the EFFECTIVE variant (`conservative`) on the decision span's `agent.prompt_variant`, so a misconfiguration is safe AND visible in the audit trail. Previously the span recorded the raw flag value, which would have hidden the fallback.

## Acceptance criteria

- [x] All three variants implemented for the LLM path, each a different prompt that shapes what the model is asked to do (`TestVariantsAreMateriallyDifferent`, goldens)
- [x] Variant selection read from the flag on EACH decision, not cached (`TestLLMDecide_VariantReachesInferencePrompt`)
- [x] Flipping the flag observably changes the prompt/decision behavior on the next decision (same test — build A vs B, System prompts differ)
- [x] Active variant recorded on the decision span as the EFFECTIVE variant; unknown → conservative fallback never dispatches an undefined prompt (`TestLLMDecide_UnknownVariantFallsBackSafely`, `TestLLMDecide_KnownVariantRecordedOnSpan`, `TestResolveVariant`)

## Out of scope (per #740)
Backend transport (#738 Jetson / #742 cloud). Async inference (#917 — stacks on this next). Anything beyond the three named variants.

## Tests
`gofmt -l .` clean, `go vet ./...`, `go build ./...`, `go test -race ./...` all green; `make lint` (ruff, no Python changes) passes.

### DIFF
--- a/services/agent/decision/layerstate.go
+++ b/services/agent/decision/layerstate.go
@@ -1,6 +1,9 @@
 package decision
 
-import "github.com/joustmania/agent/flags"
+import (
+	"github.com/joustmania/agent/flags"
+	"github.com/joustmania/agent/llm"
+)
 
 // BlockReason identifies which layer/policy blocked a decision. It is the
 // structured attribution issue #729 lifts onto the decision span.
@@ -109,11 +112,17 @@ type LayerState struct {
 // are filled in as decisions are processed.
 func newLayerState(s flags.Snapshot) LayerState {
 	return LayerState{
-		Enabled:       s.Enabled,
-		Mode:          s.Mode,
-		Objectives:    s.Objectives,
-		Model:         s.Capability.Model,
-		PromptVariant: s.Capability.PromptVariant,
+		Enabled:    s.Enabled,
+		Mode:       s.Mode,
+		Objectives: s.Objectives,
+		Model:      s.Capability.Model,
+		// Record the EFFECTIVE prompt variant (#740), not the raw flag value: the
+		// llm path resolves an unknown/garbage prompt_variant to "conservative"
+		// before building the prompt, so the decision span's agent.prompt_variant
+		// must report the variant that ACTUALLY ran — otherwise a misconfiguration
+		// would be invisible in the audit trail. ResolveVariant is the SINGLE
+		// fallback point shared with llm.Build, so span and prompt never disagree.
+		PromptVariant: llm.ResolveVariant(s.Capability.PromptVariant),
 		// inference.used defaults to rules (#741): a cycle that never resolves a
 		// backend (rules mode, or a gate-denied llm cycle) honestly reports rules. The
 		// loop overwrites this only on a gate-admitted llm cycle with the resolved tier.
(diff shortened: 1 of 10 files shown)


aepfli:
Sub-agent review: **no blockers, mergeable as-is**. Verified against a clean `go test -race ./...`: the unknown/empty variant always falls back to `conservative` (never an undefined prompt); `ResolveVariant` is the single source of truth shared by the prompt builder, the capture span, and the decision span (they can't disagree); the resolved variant reaches the REAL `backend.Infer` call (asserted via the fake's `lastPromptSystem`, not just the capture span); variant is read live per-cycle with nothing memoized; objective weights + allow-list survive in all three variants; the 6 goldens changed ONLY the VARIANT block; and the aggressive wording stays within the safety envelope (only references allow-listed interventions, and #739's `evaluatePermission` remains authoritative downstream regardless of wording).

One optional, pre-existing nit: `decision.go`'s `agent.evaluate` log line records the RAW `prompt_variant` while the decision span records the RESOLVED one. Leaving as-is — "what was configured" (log) vs "what ran" (span) is a legitimate distinction, and it's not worth churning a clean mergeable PR. Branch is already rebased onto `origin/main` (diff is #740-only).

Issue #927, opened 2026-06-10
Agent: hot-reload lifecycle/decision calibration flags (currently read-at-startup)


aepfli:
## Context
`lifecycle.player_ttl_seconds`, `lifecycle.session_grace_seconds`, `lifecycle.evict_interval_seconds`, and `decision.throttle_seconds` are read once at agent startup (`main.go` ~229–235) and never re-evaluated. This quietly contradicts the project claim that agent behavior is runtime-configurable — retuning perception/cadence requires an agent restart.

Low priority, but worth doing before the demo (#743) where live retuning is part of the story; throttle cadence also interacts with LLM call gating (#847).

## Acceptance
- These flags re-evaluated live (per-cycle or on a slow refresh tick), consistent with how `enabled`/`mode`/`objectives` already are
- Test: change throttle flag mid-session → cadence changes without restart

From the 2026-06-10 agent-layer gap evaluation.

aepfli:
we can utilize the openfeature listener for configuration changed for this

aepfli:
Delivered in #948 (merged to main): the four calibration flags (player_ttl / session_grace / evict_interval / decision_throttle) now hot-reload via the OpenFeature ProviderConfigChange listener — exactly the mechanism requested in the issue comment — into an atomic holder the throttle/eviction-ticker/store-TTLs read live, no restart. Reviewed clean. Closing per the manual-close convention.

PR #948, opened 2026-06-12
feat(agent): hot-reload lifecycle/decision calibration flags (#927)

aepfli:
## What

Part of #927. The four calibration flags were read **once at startup** (#766 F5) and baked into the store TTLs, the eviction ticker, and the decision-loop throttle, so retuning the agent's perception/cadence required a restart:

- `lifecycle.player_ttl_seconds`
- `lifecycle.session_grace_seconds`
- `lifecycle.evict_interval_seconds`
- `decision.throttle_seconds`

This PR makes them update **live with no restart**.

## How — the maintainer's requested mechanism

I implemented exactly the approach @aepfli asked for in the issue comment:

> we can utilize the openfeature listener for configuration changed for this

A domain-scoped **OpenFeature configuration-change event listener** (`openfeature.ProviderConfigChange`, registered via the agent client's `AddHandler`) drives the reload. flagd emits a config-change when its flag source changes, the Go SDK surfaces it as `ProviderConfigChange`, and the handler re-evaluates the four flags into a new concurrency-safe **`flags.LifecycleHolder`** (`atomic.Pointer`, lock-free reads / single atomic store).

- The holder is **primed once at startup** so values are correct before any event arrives (and if one never fires).
- Consumers read the holder's live source funcs on their hot path — **no per-cycle flagd polling** (a holder read is a lock-free atomic load; only the rare config-change event re-reads flagd):
  - **Decision throttle** → `Loop.shouldLog()` reads a `throttleSource func() time.Duration` (acceptance-test target).
  - **Evict interval** → `main.go`'s eviction ticker re-reads the interval each loop and `ticker.Reset`s when it changes.
  - **Player TTL / session grace** → the gamecontext `Store` reads them **at eviction time** via live source funcs; the evaluate gate's freshness window hot-reloads in lockstep.
- Existing **safe-default / non-positive guards** are preserved on every reload (a down control plane / transient bad read keeps the flagd-schema defaults), so a config-change can never collapse a TTL to zero.

The live sources are **purely additive** (nil falls back to the static construction value), so all existing eviction / throttle / gate tests are unchanged. The other decision flags (enabled/mode/objectives/policy/fitness/llm) already update per-cycle and are untouched.

## Tests (all `-race` green)

- **Acceptance (#927):** `TestThrottleSource_HotReloadsMidSession` — flips the live throttle from 10s → 2s **mid-session** (injected clock, settable source, **no live flagd / no sleeps / no network**) and asserts the `agent.disabled` span cadence reflects the new value on the next cycle.
- `TestLifecycleHolder_*` — priming, defaults-when-unset, `Reload` applies new values, a **synthetic config-change event** invokes the captured handler, and a **concurrent reload-while-reading** `-race` test (handler goroutine vs consumer reads).
- `TestStore_LiveTTLSourceHotReloads` / `TestStore_LiveSessionGraceSourceHotReloads` + nil/non-positive fallbacks — eviction reflects the live TTLs after the holder updates.
- No regression: existing gamecontext eviction, decision throttle, and #847 gate tests still pass.

## Out of scope

Anything beyond these four flags. The decision flags already update per-cycle and are not touched.

### DIFF
--- a/services/agent/README.md
+++ b/services/agent/README.md
@@ -893,17 +893,27 @@ gates, reaches the real `Writer`, and completes the `agent.action` span — all
 without touching the game. Revert by flipping `interventions_allowed` back to
 `ambient` (or `none`).
 
-### Lifecycle flags — read at startup (#766 F5)
+### Lifecycle flags — hot-reloaded via OpenFeature config-change (#766 F5, #927)
 
 The agent's lifecycle and throttle calibration values live in the flagd `agent`
-domain ([`services/flagd/agent.json`](../flagd/agent.json)). Unlike the four
-decision-cycle flag layers above (evaluated **every cycle**, hot-reloadable),
-these are **read once at startup**: they configure the GameContext store TTLs,
-the eviction ticker, and the decision-loop log/span throttle, all of which are
-fixed at construction. **Changing one requires an agent restart** — it is *not*
-hot-reload by design (per #766 F5). If flagd is not yet reachable at startup each
-value falls back to its safe default, which reproduces the former hardcoded
-constant exactly (promotion is behavior-neutral).
+domain ([`services/flagd/agent.json`](../flagd/agent.json)). Like the four
+decision-cycle flag layers above they are **hot-reloadable** — but via a
+different mechanism. #766 F5 originally read them once at startup, so retuning
+the store TTLs / eviction cadence / throttle needed a restart. **#927 makes them
+live** by wiring the **OpenFeature provider's configuration-change event
+listener** (`openfeature.ProviderConfigChange`, the mechanism the maintainer
+requested): flagd emits a config-change when its source changes, the Go SDK
+surfaces the event, and a domain-scoped handler re-evaluates these four flags
+into a shared, concurrency-safe `flags.LifecycleHolder`
+([`flags/holder.go`](flags/holder.go)). The consumers read the holder on their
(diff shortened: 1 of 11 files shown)


aepfli:
Sub-agent review: **approve, mergeable as-is**, no blockers. Confirmed the maintainer's required mechanism is genuinely implemented — a domain-scoped `client.AddHandler(openfeature.ProviderConfigChange, …)` (verified against go-sdk source: correct `*func(EventDetails)` callback-pointer lifetime, retained by the event executor, no spurious double-fire on registration), with the hot path (throttle / eviction / gate freshness) reading a lock-free `atomic.Pointer.Load()` — no polling, no per-cycle flagd call. `-race` clean; safe-default clamping airtight (no zero/negative duration can reach the ticker).

Addressed two non-blocking items in a follow-up commit:
1. The event reload now uses `context.Background()` instead of the startup SIGTERM `signal.NotifyContext` — a config-change during shutdown would otherwise reload with a cancelled ctx and briefly revert the live values to defaults. (`ctx` param dropped from `RegisterConfigChangeHandler`.)
2. Documented that the SDK dispatches each handler on its own goroutine, so concurrent `Reload` is possible — safe via the single atomic store; the delta-log is best-effort.

The other two nits (discarded `RemoveHandler` return relying on `Shutdown`; the implicit positive-ticker invariant from `durationFlag` clamping) are left as-is — both correct, just noted. `gofmt`/`go vet`/`go test -race ./...` all green.

PR #949, opened 2026-06-12
feat(agent): async LLM inference — latency budget and stale-decision application (#917)

aepfli:
Part of #917.

Makes LLM inference (#739) **asynchronous**: the per-game decision loop fires the 2–10s `Infer` in its own goroutine and returns promptly (never blocks), then re-validates the result against the **current** context before applying it — dropping stale results with a recorded reason and bounding every call with a latency budget.

## What changed
- **`decide()` fires async** (`fireInferAsync`) and returns `nil` on an admitted + reachable-tier cycle. The loop never awaits `Infer`. When async is **not** wired (no context provider) it keeps the synchronous #739 path for behavior-compatibility, so all existing #739/#741/#740 tests are unchanged.
- **`async_infer.go` / `async_apply.go`** — fire + apply-when-done machinery. On completion the result is **re-validated against the LATEST context** (fetched fresh from the per-game `gamecontext.Store` via a `ContextProvider`, not the fire-time snapshot) and re-run through the **same** permission chain (`evaluatePermission`: allow-list + battery + weighted rate limit) at **apply** time.
- **Discard reasons** (`inference.discarded_reason` on the audit span): `stale_context` / `game_ended` / `target_gone` / `permission_revoked` / `rate_limited` / `timeout`.
- **Latency budget** — new hot-tunable flag `llm.latency_budget_seconds` (default 8s, in `LLMGate`, mirroring the #847 gate flags in `services/flagd/agent.json` + `flags/flags.go`). Each call is bounded by `context.WithTimeout`; an over-budget result is dropped and the deterministic rules engine decides for the current context (#741 timeout → rules), so the system always decides.
- **Pile-up guard** — an atomic in-flight flag per `Loop` ensures at most **one** `Infer` per game at a time. A cycle that finds one in flight skips firing (`inference.fallback_reason=llm_inflight`) and falls back to rules.
- **Audit** — a dedicated `agent.llm.apply` span records `inference.latency_ms` and the applied/discarded outcome (with the discard reason); the `agent.llm.infer` call span hangs off it with the real call timestamps.
- **Budget coherence (#847)** — the global `llmBudget` is charged at **fire** time (admission, in `gateLLM`); the per-game weighted rate limiter is charged at **apply** time through `evaluatePermission`. Neither double-charged nor skipped.
- **Leak-free** — every fired goroutine is bounded by the agent root context and tracked on a `WaitGroup`; `LoopSet.AwaitInflight` joins them at graceful shutdown.

## Design fork resolved (please review)
The issue specifies behavior but not whether rules runs as a stopgap each cycle. Two readings of "loop never blocks" + "timeout → rules" + "system always decides":

- **(A)** run rules immediately every admitted cycle and apply the async LLM result later as an *additional* decision; or
- **(B)** the admitted cycle decides nothing synchronously and the single decision for that context is produced by the async path — the model's (re-validated + applied) or, on timeout/error/unparseable, the rules engine **in the async completion path** for that same context.

I implemented **(B)**. It matches #741's fallback chain (timeout/error/unparseable → rules is a property of the inference *attempt*, not a parallel rules cycle) and yields **one** decision per admitted context — (A) would dispatch two interventions per admitted cycle and double-charge the rate limiter. The loop still never blocks; the system still always decides, a beat later. Rationale is documented in `async_infer.go`'s header.

## Tests
All in `decision/async_infer_test.go`, with fake backends (channel-blocking / hanging / erroring / ctx-honoring) and a fake context provider that flips state between fire and apply — no real sleeps, no network:
- loop returns while `Infer` is still running (never blocks);
- stale results provably discarded (game ended / target gone / partition evicted) — not applied, span carries the right `discarded_reason`;
- a fresh result IS applied through the permission chain;
- apply-time re-validation blocks (permission_revoked via battery, rate_limited);
- latency-budget timeout drops the result and falls back to rules, span records the timeout + latency;
- pile-up guard: a second cycle while one is in flight does not launch a second `Infer`;
- infer-error falls back to rules.

Plus `flags/flags_test.go` coverage for the new latency-budget flag (flagd shape, default, non-positive floor).

`gofmt -l .` clean, `go vet ./...` clean, `go build ./...`, `go test -race ./...` all green (run ×3 for flakiness). `make lint` (ruff) passes.

## Out of scope (unchanged)
Real Jetson/cloud transport (#738/#742 — `Infer` stays the sentinel stub in prod; async is exercised via fakes); significance-triggered escalation; gate/resolver/variant semantics.

### DIFF
--- a/services/agent/decision/async_apply.go
+++ b/services/agent/decision/async_apply.go
@@ -39,8 +39,9 @@ import (
 //   - applies it through the SAME runDecision permission chain a sync decision uses.
 //
 // On timeout / infer-error / unparseable it falls back to the deterministic rules
-// engine FOR THE CURRENT CONTEXT, so the game always gets a decision — the #741
-// timeout->rules chain, honoring "the system always decides".
+// engine FOR THE CURRENT CONTEXT — the #741 timeout->rules chain, so the rules
+// engine is always CONSULTED (subject to its own evalInterval cadence; see
+// fallbackToRules) rather than the loop silently dropping the cycle.
 func (l *Loop) applyAsyncResult(root context.Context, snapshot flags.Snapshot, trig EvalTrigger, out asyncOutcome) {
 	// Build the apply audit root. It is backdated to the trigger T0 (the fire cycle's
 	// Export arrival) so the trace duration covers the whole fire->infer->apply span,
@@ -206,8 +207,12 @@ func (l *Loop) applyOneAsync(ctx context.Context, applyRoot trace.Span, snapshot
 // fallbackToRules runs the deterministic rules engine for the CURRENT context and
 // applies its decisions when the async LLM result could not be used (#917): a
 // timeout, or an infer-error / unparseable response. This is the #741
-// timeout->rules chain executed in the async completion path, so the game always
-// gets a decision even though the loop never blocked.
+// timeout->rules chain executed in the async completion path, so the rules engine
+// is CONSULTED for the current context even though the loop never blocked. Note it
+// is "consulted", not "always emits": ObjectiveRules carries its own ~1s
+// evalInterval throttle, so if the synchronous loop ran rules <1s before this
+// timeout lands, Evaluate returns no decision this round — the engine's intended
+// cadence, not a dropped decision.
 //
 // cause is what made the LLM result unusable: DiscardTimeout (budget elapsed,
 // recorded as the discard reason) or FallbackUnparseable (the tier answered, but
--- a/services/agent/decision/async_apply.go
+++ b/services/agent/decision/async_apply.go
@@ -1,0 +1,303 @@
+package decision
+
+import (
+	"context"
+	"time"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/joustmania/agent/flags"
+	"github.com/joustmania/agent/gamecontext"
+	"github.com/joustmania/agent/llm"
+)
+
+// async_apply.go is the APPLY-when-done half of async inference (#917): it runs in
+// the fired goroutine when Infer returns (or times out), re-validates the result
+// against the CURRENT game context, and applies or discards it — recording the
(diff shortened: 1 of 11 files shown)


aepfli:
Sub-agent review: **approve, mergeable as-is**, zero blockers — and this was the most invasive PR in the stack. The reviewer ran `go test -race -count=1 ./...` (all 9 packages green) and traced every concern: the in-flight guard release is airtight (defers fire on panic, no claim-without-release path), no goroutine leak (`GracefulStop` drains Export RPCs before `AwaitInflight`, so no Add-after-Wait), re-validation re-fetches the LIVE context (all six discard reasons map correctly), permission parity via the same `runDecision`/`evaluatePermission` chain, and no double-charge (global budget at fire, weighted limiter once at apply, neither on discard).

Addressed the comment-level nits in a follow-up commit: softened the "system always decides" wording on the timeout path (rules are *consulted*, subject to ObjectiveRules' own ~1s cadence) and documented that evicting a loop with an in-flight inference self-heals (bounded goroutine → stale_context discard). The remaining notes (pile-up double-intervention, fired cycle has no `agent.span_received` root, mux-by-reference) are acknowledged-correct observability points, left as-is.

Note for merge ordering: #948 (#927) edits the throttle area of `decision.go` while this PR adds `+trig` to `decide()` and the llm branch — different regions, no semantic interaction, but a textual conflict is possible; whichever merges second I'll rebase.

aepfli:
Rebased onto current main after #948 (#927) and #950 (M7-1) merged. Single conflict in `main.go`: this PR declares `var mux` earlier (so the per-game loop factory's `MuxContextProvider` can close over it) and assigns `mux = NewMultiplexer(...)`, while main had `mux := ...` with #927's live-TTL sources + #950's `chainGameEnd` summary wiring. Resolved by keeping the #927/#950 factory body and using this PR's `mux =` assignment form. `gofmt`/`go vet`/`go build`/`go test -race ./...` all green; diff is #917-only. `MERGEABLE` again.

Issue #933, opened 2026-06-12
M7-6: Dashboard intelligence pane


aepfli:
## M7-6: Dashboard intelligence pane

The live demo dashboard gains an "intelligence" pane that surfaces what the agent is thinking and how synthetic-game quality is trending.

### What needs to be built
A live dashboard pane that displays, updating in real time during a session:
- The **last LLM decision** together with the **cited signal** that drove it.
- The current **simulation quality score**.
- **Mock improvement history** (agent-initiated changes to the simulation/mock over time, incl. PRs opened / changes reverted).
- Any **agent-initiated game interventions**.
- The current code-improvement **mode** (`issue | pr | autonomous`) and **allowed_paths** scope, reflecting live flag values.
- For **synthetic games**, a **quality-score badge** shown alongside the existing `SYNTHETIC` indicator.

### Acceptance criteria
- [ ] All elements (last decision + cited signal, simulation quality score, mock improvement history, agent-initiated interventions, current mode + allowed_paths) are visible during a live session.
- [ ] Synthetic games show a quality-score badge next to the SYNTHETIC indicator.
- [ ] Mode/scope reflect live flag values and update instantly on flag flip.
- [ ] All elements update in real time.

### Dependencies
- Depends on **M5 dashboard** (the demo dashboard from M5 Demo Polish — Agent tab #792).
- Depends on **M7-1** (#928) — game narrative builder (supplies the quality signal).

### Labels
`demo`

aepfli:
Moved from M7 to **M5 Demo Polish**: this is the demo dashboard's intelligence PANE (presentation), not the intelligence/self-improvement LOGIC (which is built + merged + closed: #928/#929/#930/#931/#932/#935/#936). It depends on the M5 dashboard / Agent tab (#792) where it's rendered, and the live data it surfaces (LLM decisions, simulation-quality scores, improvement history / PRs opened-reverted, live mode + allowed_paths) only exists once the M8 live-wiring lands (#964/#965). So its natural home is M5 (the dashboard), gated on M8 for real data. M7 (logic) is complete.

aepfli:
Folded into #792 (Agent focus tab) — the intelligence pane is more panels on the same tab, and its live data is M8-gated. Scope captured on #792. M7 logic is complete; this is its presentation surface.

Issue #930, opened 2026-06-12
M7-3: Prompt context injection via flags


aepfli:
## M8-3: Prompt context injection via flags

An operator can append runtime context to the agent's system prompt through a free-text feature flag, without redeploying or touching base game facts.

### What needs to be built
- A free-text OpenFeature flag `agent.prompt.context_note` whose value is appended to the agent's system prompt at inference time.
- The injected context must be **validated and length-bounded** before it reaches the model:
  - Reject **oversized** values (enforce a maximum length).
  - Reject **invalid** values (e.g. empty/malformed) before they are sent to the LLM.
- Injected context is visible as a span attribute on **every LLM call**.
- **Base game facts remain hardcoded** — the flag only *appends* operator context, it does not replace the authoritative prompt foundation.

### Acceptance criteria
- [ ] The `agent.prompt.context_note` flag value appears in the LLM prompt when set.
- [ ] The injected context is visible as a span attribute on every LLM call.
- [ ] Invalid or oversized values are rejected before reaching the model.
- [ ] Base game facts remain hardcoded and are not overridable via the flag.

### Labels
`agent` `openfeature` `inference`

aepfli:
Delivered in #952 (merged to main / M7-3): the `prompt.context_note` operator free-text flag, read live each cycle, validated + length-bounded (`ValidateContextNote`: rejects empty/oversized >1024 runes/control chars, never truncates, CRLF-normalized), injected as a non-authoritative OPERATOR CONTEXT section after the hardcoded base facts (base facts stay authoritative; allow-list gate remains the real backstop), with `context_note_present`/`context_note_len` on every LLM call (sync + async). Reviewed clean. All 4 ACs met. Closing per the manual-close convention.

Issue #929, opened 2026-06-12
M7-2: Rolling game context window


aepfli:
## M7-2: Rolling game context window

The agent assembles the last N game summaries into a narrative context block that is fed into every LLM call, giving the model memory across sessions.

### What needs to be built
- Maintain a rolling window of the most recent **N** game summaries produced by the narrative builder (M7-1, #928).
- Assemble them into a single **narrative context block** that is injected into every `agent.inference` LLM call.
- **N is flag-controlled** via OpenFeature (runtime-tunable, no redeploy).
- The context block updates after **every completed game** (the window advances).
- The window size is recorded as a span attribute `agent.llm.context_games = N` on every `agent.inference` span.

### Acceptance criteria
- [ ] Each LLM call includes the last N game summaries as a narrative context block.
- [ ] `agent.llm.context_games = N` is visible on the `agent.inference` span in the trace.
- [ ] The context updates after every completed game.
- [ ] N is controlled by a feature flag and tunable at runtime.

### Dependencies
- Depends on **M7-1** (#928) — game narrative builder.

### Labels
`agent` `inference` `openfeature`

aepfli:
Delivered in #951 (merged to main / M7-2): a global bounded ring of the last N game summaries injected as a 'PRIOR GAMES' narrative block into every LLM prompt, N driven live by the `llm.context_games` flag, with the injected count on the inference + capture spans. Wired into BOTH the sync and async (#917) prompt paths. Reviewed clean (including a recovery from a session-cut-off first pass + the async re-integration after #917 merged). Closing per the manual-close convention.

Issue #928, opened 2026-06-12
M7-1: Game narrative builder


aepfli:
## M7-1: Game narrative builder

After each game, the agent aggregates raw spans into a compact game summary so it can reason about what happened rather than re-reading low-level telemetry.

### What needs to be built
- A post-game aggregation step that consumes the raw spans for a completed game and produces a compact **game summary**.
- The summary must capture the key signals that describe the game as a narrative:
  - **Player arcs** — per-player trajectory across the game (joined, eliminations, survival time).
  - **Engagement trend** — how activity/intensity evolved over the game.
  - **Elimination spread** — distribution/timing of eliminations.
  - **Dominant player detection** — whether one player ran away with the game.
  - **Intervention outcomes** — which agent interventions fired and their before/after effect.
- Summaries are stored as **JSON on disk**, one file per game.
- Runs after **both real and synthetic games**.
- The aggregation itself is **visible as a span** in the trace.

### Acceptance criteria
- [ ] After every real and synthetic game a summary file is written to disk.
- [ ] The summary contains all key signals: player arcs, engagement trend, elimination spread, dominant player detection, intervention outcomes.
- [ ] The aggregation is visible as a span in the trace.

### Dependencies
- Depends on **M6 Shadow Games** (milestone done) — the shadow-game runner (#778) and multi-session coordinator (#775) already produce synthetic games, so the builder runs after both real and synthetic games.

### Labels
`agent` `otel` `gamestack`

aepfli:
Delivered in #950 (merged to main / M7-1): after every real and shadow game the agent aggregates the pre-reset GameContext (live signals + the #916 timeline) into a compact JSON Summary on disk (player arcs, engagement trend, elimination spread, dominance, intervention outcomes), emitted as an agent.game.summary span. Reviewed clean. Note: intervention outcomes stay latent until #945 wires Store.AppendInterventionEvent. Closing per the manual-close convention.

PR #950, opened 2026-06-12
feat(agent): game narrative builder — per-game JSON summaries (#928)

aepfli:
Part of #928 (M7-1: Game narrative builder).

## What

After each game ends — **real AND shadow** — aggregate the partition's pre-reset `GameContext` snapshot (live signals + the #916 rolling timeline) into a compact **game summary** so the agent can reason about *what happened* as a narrative instead of re-reading low-level telemetry.

New `services/agent/gamesummary` package:

- **`summary.go`** — pure, clock-free `BuildSummary(GameContext, BuildOptions) Summary`. Table-testable, deterministic over its input (only `GeneratedAt` is injected; the aggregation never reads a clock/network/disk).
- **`writer.go`** — atomic JSON writer (temp file + `fsync` + `rename`), **one file per `game_id`**, dir from `AGENT_GAME_SUMMARY_DIR` (sensible default, created if missing). `game_id` sanitized against path traversal.
- **`coordinator.go`** — once-per-game `OnGameEnd` hook (bounded LRU dedupe, mirroring the #844 retro) that builds + writes the summary under an **`agent.game.summary`** span carrying low-cardinality counts.

## Summary schema (JSON)

`schema_version`, `game_id`, `game_kind`, `game_mode`, `generated_at`, `duration_seconds`, `player_count`, `elimination_count`, `intervention_count`, plus the five required signals:

- **`player_arcs`** — per-serial: `eliminated`, `elimination_order`, join/elimination offsets, `survival_seconds`, `final_skill`/`final_intensity`.
- **`engagement_trend`** — oldest-first `(offset, active_players, mean_intensity)` series.
- **`elimination_spread`** — ordered `(offset, serial, order)`.
- **`dominance`** — `dominated` flag + `serial` + named `heuristic` + `margin`.
- **`interventions`** — `(intervention, target, dispatched, offset, effect)`.

## Signal provenance

Derived **from the #916 timeline**: engagement trend (state-delta entries), elimination spread + per-player order (elimination entries), intervention outcomes (intervention entries). Player arcs fold live `PlayerSignals` with the timeline; dominance is a coarse, explainable `survivor_outlasted_mean_2x` heuristic.

**Intervention before/after effect** is a best-effort *correlation*: mean-intensity and active-count delta between the 30 s timeline windows immediately before vs after each intervention's timestamp (`nil` when neither side has a comparable sample). The intervention stream is the timeline's `EventIntervention` entries (`Store.AppendInterventionEvent`); that seam is currently unwired in production (see the #916 deferral on `AppendInterventionEvent`), so the list is usually empty until the seam is threaded — the aggregator handles them correctly the moment they appear (and the tests drive them).

## Wiring / double-fire avoidance

The store exposes a single `OnGameEnd` hook. The summary coordinator is chained alongside the #844 retro via a small `chainGameEnd` helper in `main.go`. Each consumer owns its **own** once-per-game dedupe, so chaining cannot make either double-fire. The summary is a **separate** component (no LLM budget / eligibility gate) so it runs unconditionally for every game — a budget-exhausted game still gets a summary.

## Tests

- `BuildSummary` table tests: player arcs, elimination spread, engagement trend, dominance (detected + not), intervention outcomes + effect delta, empty timeline, determinism.
- Writer: file-per-game, dir-created-if-missing, valid JSON round-trip, **atomic / no partial file / no `.tmp` leftover**, `game_id` sanitization, concurrent writes.
- Coordinator: file + span emitted, runs for **shadow** games, exactly-once per game, skips non-end snapshots, two concurrent games → two files, end-to-end driven from a real `Store`.

`gofmt -l` clean, `go vet ./...` clean, `go test -race ./...` all green. No new dependencies (`go.sum` untouched).

## Out of scope (later M7)

Multi-game rolling **window** (#929 M7-2 — this is ONE summary per game), prompt injection (#930), dashboard pane (#933), and deep causal before/after attribution beyond the timeline-window delta.

### DIFF
--- a/services/agent/README.md
+++ b/services/agent/README.md
@@ -857,6 +857,7 @@ All configuration is via environment variables:
 | `AGENT_PROBE_DECISIONS` | _unset_ | `true` enables the demo/verification probe: a synthetic `noop` decision (and thus a full audit trace) at most every 5 s. Never for production sessions. See [Probe mode](#probe-mode-agent_probe_decisions) |
 | `AGENT_INTERVENTIONS_ENABLED` | `false` | `true` swaps the no-op action sink for the real intervention **Writer** (#730). Default off keeps the scaffold inert |
 | `INTERVENTIONS_FLAG_PATH` | `/etc/flagd/interventions.json` | Path of the flagd interventions file the Writer rewrites (must be the bind-mounted file flagd watches) |
+| `AGENT_GAME_SUMMARY_DIR` | `/var/lib/joustmania/agent/summaries` | Directory the M7-1 game narrative builder (#928) writes one JSON game summary per game into (real **and** shadow), created if missing; atomic temp+rename so a reader never sees a partial file |
 
 > The Go agent uses the flagd **RPC** resolver (gRPC evaluation port `8013`),
 > not the in-process sync port `8015` that the Python services use.
--- a/services/agent/gamesummary/coordinator.go
+++ b/services/agent/gamesummary/coordinator.go
@@ -1,0 +1,183 @@
+package gamesummary
+
+import (
+	"context"
+	"log/slog"
+	"sync"
+	"time"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/joustmania/agent/gamecontext"
+)
+
+// coordinator.go is the M7-1 game-end glue (#928): it hangs off the same
+// gamecontext.Store.OnGameEnd signal the #844 retrospective uses, builds the
+// Summary from the pre-reset snapshot, writes it to disk, and records the
+// aggregation as an agent.game.summary span. It fires EXACTLY ONCE per game for
+// BOTH real and shadow games.
+//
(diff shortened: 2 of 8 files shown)


aepfli:
Sub-agent review: **approve, no blockers** — verified no double-fire with the #844 retro (both take `GameContext` by value, each owns its dedupe, summary is un-gated so it fires for every real + shadow game), no panic/div-by-zero on edge games (empty timeline / 0 players / all-eliminated / single player all guarded), atomic temp-file+rename write with `safeFilename` path-traversal protection (`../../etc/evil` proven contained), and `-race` clean across all packages.

Addressed the one real finding in a follow-up commit: `PlayerArc.JoinOffsetSeconds` was a **dead field** — documented in the struct/package/PR but never assigned, because state-delta timeline events are session-scoped (no serial), so a surviving player has no join anchor. Dropped the field + corrected the docs rather than fake it (a truthful join offset needs Store-level per-player join timestamps — a separate change). Also cleared a stale `GameStartOffset` doc reference and a dangling `_ = dir` test line. The remaining nit (redundant nil-`now` guard) is harmless defensive code, left as-is.

The intervention signal is honestly latent until #945 wires `Store.AppendInterventionEvent` (degrades to an empty list, effect is named correlation-not-causation) — the aggregator + tests already cover it. `gofmt`/`go vet`/`go test -race` green.

PR #951, opened 2026-06-12
feat(agent): M7-2 rolling game context window (#929)

aepfli:
## What

M7-2: a **rolling cross-game context window** (#929) for the Go agent. The agent now carries memory across games: every LLM decision prompt includes the last N recent game summaries as a compact narrative context block, so the model can reason about *what happened in the last few games* — not just the current game's live snapshot + the #916 in-game timeline.

This PR was **finished from a cut-off first pass**: the foundational pieces were already committed as a WIP commit (the `gamewindow` package, the `llm.context_games` flag, the `AttrLLMContextGames` span const, the gamesummary `Observer` seam, and `BuildInput.ContextBlock`). This change **wires them together and adds the full test suite**. The two commits squash on merge.

## Feature pieces

- **Window** (`gamewindow/`): `Store` is a bounded, mutex-guarded ring (`RetentionCap=20`) of recent `gamesummary.Summary` values; `Record` appends + evicts oldest, `Recent(n)` hands back a fresh copy of the last n (oldest-first). `Render` produces a deterministic, compact `PRIOR GAMES` block (header counts + dominance + intervention outcomes), `(no prior games)` when empty.
- **Flag** (`flags.go` + `services/flagd/agent.json`): `llm.context_games` read **live** each cycle (`Snapshot.ContextGames`, default 3, tight/patient variants), so N is retunable with no redeploy.
- **Window advance** (`gamesummary/coordinator.go` → `main.go`): one global `gamewindow.Store` is registered as the coordinator's `Observer`, so it `Record`s each freshly-built summary on game end (before the disk write), for both real and shadow games.
- **Injection** (`decision/llm_decide.go`): on the LLM path the loop reads `ContextGames`, clamps to `[0, RetentionCap]`, pulls `Recent(N)`, renders, and passes the block as `BuildInput.ContextBlock`. The same block is injected on the capture path (`agent.llm.prompt`).
- **Span** (`telemetry.go`): `agent.llm.context_games` records the **count actually injected** (`len(Recent)`, bounded by games-ended) — not the raw flag — on both the `agent.llm.infer` and `agent.llm.prompt` spans.
- One global `Store` shared across all per-game loops (`SetContextWindow`), exactly like the shared LLM budget/resolver. **Nil-safe**: a loop with no window injected renders no cross-game block and reports count 0 — purely additive over the #739 path.

## Where the block + span land in the LLM path

[...]

Issue #931, opened 2026-06-12
M7-4: Code improvement via remote coding agent


aepfli:
## M7-4: Code improvement via remote coding agent

When the agent identifies a problem from game narratives, it invokes a **remote coding agent** to propose a validated code change within an allowed scope. No Pi, no Jetson — the coding agent runs as a side process backed by **Ollama on a separate laptop** (the same side inference node the rest of the stack uses).

### What needs to be built
- The agent has access to the Tier 2 codebase scope defined in `AGENTS.md` (relevant source files, fitness-function definitions, improvement boundary).
- When a game-narrative signal indicates a Tier 2 problem — **simulation quality**, **game balance**, **coordinator behaviour** — the agent invokes a **remote coding agent** via its SDK/RPC interface, passing structured context: the implicated source file, the fitness/narrative signal, and a hypothesis about the cause.
- The coding agent is backed by the side **Ollama** inference node (laptop on the side) — no cloud dependency required, consistent with the inference stack. The backend is flag-selectable via `agent.code_improvement.engine`, so it can change without code edits.
- The coding agent returns a **proposed diff** referencing a specific file within the allowed scope (boundary enforced by M7-5, #932).
- The proposal — context, hypothesis, diff summary, full reasoning — is logged as a trace span **before any action is taken** (`code_improvement.proposed_diff_summary`).
- The proposed change is validated through the synthetic suite (M7-7) and promoted through the mode flow (M7-8); those stages are separate issues.

### Acceptance criteria
- [ ] Agent invokes a remote coding agent via SDK/RPC with structured context (file + signal + hypothesis).
- [ ] The coding agent is backed by the side Ollama node; no cloud dependency in the path.
- [ ] At least one proposed change is derived from a game-narrative signal and references a specific file in allowed scope.
- [ ] Proposed diff summary + full reasoning are visible as a span (`code_improvement.proposed_diff_summary`).
- [ ] Engine backend is flag-selectable via `agent.code_improvement.engine`.

### Dependencies
- Depends on **M7-2** (#929) — rolling game context window (supplies the narrative context).
- Gated by **M7-5** (#932) — allowed_paths. Validated by **M7-7** (synthetic gate). Promoted by **M7-8** (mode flow).

### Labels
`agent` `inference` `gamestack`

aepfli:
## Refined design — agreed with maintainer 2026-06-12 (supersedes the original body)

[...]

PR #952, opened 2026-06-12
feat(agent): operator prompt context note via flag (#930)

aepfli:
## M7-3: Prompt context injection via flags

An operator appends **free-text runtime context** to the agent's system prompt via the `prompt.context_note` OpenFeature flag — **no redeploy**, and **without touching the authoritative base game facts**. The note is validated + length-bounded before it reaches the model and is visible as a span attribute on every LLM call.

> **Stacked on #951 / M7-2 (#929), which has since MERGED to `main`.** This branch was cascade-rebased onto `origin/main` after #951's squash-merge, so the PR targets `main` directly and the diff is **#930-only**.

### What changed
- **Flag** (`flags/flags.go`, `services/flagd/agent.json`): `keyPromptContextNote = "prompt.context_note"` (flat key, matching `mode` / `llm.context_games` — the `agent.` in the issue is the flagd **domain**, not the key), `DefaultPromptContextNote = ""`, and `Snapshot.PromptContextNote` read **live every cycle** via the string getter (mirrors #929 `ContextGames`). flagd flag defaults to empty.
- **Prompt builder** (`llm/prompt.go`): `BuildInput.ContextNote` (pre-validated) injected by `buildSystem` as a clearly-delimited, explicitly **non-authoritative** `OPERATOR CONTEXT` section **after** the hardcoded base facts and the #929 `PRIOR GAMES` block. `Build` stays PURE. Empty note → no section (existing goldens unchanged).
- **Validation gate** (`llm.ValidateContextNote`, pure + table-tested): empty/whitespace → reject; control chars (all except `\n`) → reject; length > `MaxContextNoteLen` (**1024 runes**) → **reject, never truncate** (a cut could sever an instruction mid-sentence). A rejected note leaves the prompt exactly as if none were set.
- **All three prompt-build sites** wired through one `resolveContextNote` seam: sync `llmDecide`, async `runInfer` (carried on `asyncOutcome` to apply time), and `captureLLMPrompt`.
- **Span attribute on every LLM call**: bounded low-cardinality view — `agent.llm.context_note_present` (bool) + `agent.llm.context_note_len` (int, runes) on `agent.llm.infer` and `agent.llm.prompt`. We record present+len rather than the raw text: the note is operator-set (not PII) but free-text/long, so stamping it raw would balloon span cardinality.

### Safety / base-facts authority
- Base facts are **hardcoded and precede** the note; the operator section is fenced and framed as supplementary ("does NOT override the base facts…"), so a huge or hostile note can neither blow the prompt budget (length bound) nor override the foundation (ordering + framing).
- **Fail-safe**: any validation failure or flagd-unreachable → empty note → prompt identical to no-note. The decision never errors and never injects unvalidated text.

### Acceptance criteria
- [x] `prompt.context_note` value appears in the LLM prompt when set (valid + within length).
- [x] Injected note visible as a span attribute on every LLM call.
- [x] Invalid/oversized values rejected before reaching the model (oversized → not injected; empty → not injected; prompt unchanged).
- [x] Base game facts remain hardcoded, always present, and precede the note — not overridable via the flag.

### Tests
`ValidateContextNote` table (empty/normal/exactly-max/over-max/control/multibyte); prompt injection + base-precedence + clean-insert; sync, async (fire → `AwaitInflight`), and capture-path injection; live-flag flip; rejection on both sync + async paths; span present/len. Full `go test -race ./...` green; `gofmt -l` clean; `go vet ./...` clean.

### Out of scope (later M7)
Code-improvement self-improvement track (#931+) and the dashboard (#933). This is ONLY the operator free-text note; #929 (cross-game window) is the separate M7-2 feature already on the base.

Part of #930 (M7-3).

### DIFF
--- a/services/agent/decision/async_apply.go
+++ b/services/agent/decision/async_apply.go
@@ -138,6 +138,13 @@ func (l *Loop) emitAsyncInferSpan(ctx context.Context, out asyncOutcome) ([]Deci
 	// the prompt on the REAL inference span, mirroring the synchronous llmDecide path
 	// (AttrLLMContextGames). out.contextGames is what runInfer rendered (0 = un-wired).
 	span.SetAttributes(attribute.Int(AttrLLMContextGames, out.contextGames))
+	// M7-3 (#930): the operator-note view (present + rune length) on the REAL inference
+	// span, mirroring the synchronous llmDecide path. out.contextNote* is what runInfer
+	// validated + injected at fire time (present=false/len=0 = no/rejected note).
+	span.SetAttributes(
+		attribute.Bool(AttrLLMContextNotePresent, out.contextNotePresent),
+		attribute.Int(AttrLLMContextNoteLen, out.contextNoteLen),
+	)
 	defer span.End(trace.WithTimestamp(out.end))
 
 	if out.timedOut {
--- a/services/agent/decision/async_infer.go
+++ b/services/agent/decision/async_infer.go
@@ -212,11 +212,19 @@ func (l *Loop) runInfer(root context.Context, snapshot flags.Snapshot, backend B
 	// carried on the outcome and stamped on the agent.llm.infer span at apply time
 	// (emitAsyncInferSpan), mirroring how the sync path records AttrLLMContextGames.
 	contextBlock, contextCount := l.renderContextBlock(snapshot)
+	// M7-3 operator note (#930): inject the SAME validated operator context note the sync
+	// llmDecide path would, so the PRODUCTION (async) path actually carries it — #917
+	// moved prompt assembly here, so without this the feature would be bypassed in
+	// production (the same trap M7-2 hit). present/len are carried on the outcome and
+	// stamped on agent.llm.infer at apply time (emitAsyncInferSpan), mirroring the sync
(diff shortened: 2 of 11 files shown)


aepfli:
Sub-agent review: **mergeable as-is, no blockers**. Verified validation is airtight + fail-safe (oversized/control-char can't reach the model; rune-counted; never truncates; any rejection → prompt byte-identical to no-note), all three prompt-build sites including the async production path inject through one `resolveContextNote` seam (`-race` clean), span attrs are bounded (present+len), and base-facts authority holds both structurally (note appended last, delimited, framed non-overriding) AND via the authoritative post-hoc allow-list gate (#739/#946) — even a successful prompt-injection can't dispatch an out-of-allow-list action.

Applied the one usability nit in a follow-up commit: CRLF/CR line endings are now normalized to LF before the control-char scan, so an operator pasting a Windows/Mac multi-line note gets the note injected instead of a silent no-op (no safety cost — LF was already allowed, every other control byte still rejected). Table test updated. The other nit (sync test reads `lastPromptSystem` directly) is cosmetic, left as-is. `gofmt`/`go vet`/`go test -race ./...` green.

Issue #936, opened 2026-06-12
M7-8: Code-improvement promotion flow — mode flag + GitHub PR


aepfli:
## M7-8: Code-improvement promotion flow — mode flag + GitHub PR

A validated change is promoted according to a runtime mode flag: open a GitHub issue with a hypothesis, open a PR with fitness evidence, or apply autonomously and watch live games. The mode flip is the live demo moment.

### What needs to be built
- Read `agent.code_improvement.mode` (`issue | pr | autonomous`) at promotion time:
  - **issue** — open a GitHub issue describing the hypothesis and the implicated file; no code applied.
  - **pr** — open a PR with a structured, machine-readable description: fitness before/after, synthetic game results, coding-agent reasoning summary, and flag state at time of improvement.
  - **autonomous** — apply the validated change, watch real games, revert if fitness degrades (per M7-7).
- `agent.code_improvement.target` (`local | github`) selects where a PR/commit lands. When `github` and connectivity is available, commit + open a PR via the GitHub API; otherwise apply locally.
- Mode and target are observable in traces: `code_improvement.gate`, `code_improvement.target`, `code_improvement.outcome = applied | discarded | reverted`.
- Flipping the mode flag takes effect on the **next promotion without redeploy** — this is the on-stage moment (issue → autonomous, live).

**Out of scope (backlog):** the "GitHub unreachable → apply local → retroactively open PR when connectivity returns" reconnect fallback. Keep PR-open straightforward; track the offline-reconnect path separately if it's ever needed.

### Acceptance criteria
- [ ] `agent.code_improvement.mode` controls promotion: `issue` opens an issue, `pr` opens a PR, `autonomous` applies + watches.
- [ ] PR description is structured: fitness before/after, synthetic results, reasoning summary, flag state.
- [ ] `agent.code_improvement.target` selects local vs github.
- [ ] Outcome visible in traces (`code_improvement.outcome`).
- [ ] Flipping `mode` at runtime changes behaviour on the next promotion without redeploy.

### Dependencies
- Depends on **M7-7** (synthetic validation gate) — only validated changes are promoted.

### Labels
`agent` `openfeature` `infrastructure`

aepfli:
[...]

Issue #935, opened 2026-06-12
M7-7: Synthetic validation gate for code improvements


aepfli:
## M7-7: Synthetic validation gate for code improvements

Before any proposed code change is promoted, the agent validates it against the M6 synthetic game suite and compares fitness to a baseline. Only changes that demonstrably improve fitness are promoted.

### What needs to be built
- Apply the proposed diff (from M7-4, #931) to a **staging copy** of the service.
- Run the **M6 synthetic game suite** against the patched code; game count configurable via `agent.code_improvement.validation_games`.
- Compute fitness on the patched build and compare to a baseline established from the last N real games. Record `code_improvement.fitness_before` / `code_improvement.fitness_after`.
- **Promote** the change if improvement exceeds `agent.code_improvement.fitness_improvement_threshold`; **discard** otherwise.
- **Revert** on degradation when `agent.code_improvement.revert_on_degradation = true`.
- The full cycle is one trace tree: `evaluation → invocation → synthetic_validation → apply | discard`.

### Acceptance criteria
- [ ] Proposed diff applied to a staging copy; the agent runs the M6 synthetic suite against it.
- [ ] Game count controlled by `agent.code_improvement.validation_games`.
- [ ] `code_improvement.fitness_before` and `code_improvement.fitness_after` recorded on the span.
- [ ] Change promoted only if improvement exceeds `fitness_improvement_threshold`, discarded otherwise.
- [ ] Revert on degradation when `revert_on_degradation = true`.
- [ ] Full validation cycle visible as a single trace tree.

### Dependencies
- Depends on **M7-4** (#931) — supplies the proposed diff.
- Builds on **M6 Shadow Games** (milestone done) — the synthetic suite (#774–#893).

### Labels
`agent` `otel` `gamestack`

aepfli:
Delivered in #956 (merged / M7-7): the `experiment.Validator` — baseline → run synthetic → measure fitness → promote/discard/revert verdict (verdict-only; the real-default move is #936), fail-closed on seam errors, single trace tree. Built with injectable seams; the LIVE wiring to real shadow games + #731 fitness is M8 (#965). Reviewed clean. Closing per the manual-close convention.

Issue #932, opened 2026-06-12
M7-5: Allowed paths gate for code improvement


aepfli:
## M7-5: Allowed paths gate for code improvement

The agent's code-improvement invocation is constrained to a flag-controlled allow-list. Any proposed change outside the declared paths is blocked before it is applied, and the block is observable in traces.

### What needs to be built
- The `agent.code_improvement.allowed_paths` flag defines the only paths the remote coding agent (M7-4, #931) may modify, covering the Tier 2 scope required for autonomous game improvement, e.g.:

```
agent.code_improvement.allowed_paths = [
  "services/game_coordinator/games/",
  "services/mock/",
  "config/"
]
```

- The gate reads `allowed_paths` from flagd **on each invocation** (not cached), so changes take effect on the next invocation **without redeploy or restart**.
- Any proposed change touching a file outside the allow-list is **blocked before it is applied**.
- The boundary protects the agent's own decision logic, the flag schema, and the OTel instrumentation — these must never be in scope.
- Blocked attempts emit a `code_improvement.blocked = true` span attribute.

### Acceptance criteria
- [ ] The remote coding agent cannot modify files outside the declared `allowed_paths`.
- [ ] Blocked attempts produce a `code_improvement.blocked = true` span attribute.
- [ ] `allowed_paths` is read from flagd per invocation; a change takes effect on the next invocation without restart.
- [ ] An attempt to modify agent decision logic or the flag schema is blocked.

### Dependencies
- Depends on **M7-4** (#931) — code improvement via remote coding agent (this gate constrains what that invocation can touch).

### Labels
`agent` `openfeature` `infrastructure`

aepfli:
## Refined design — agreed with maintainer 2026-06-12 (supersedes the original body)

This gate is now about **flag experiments, not source edits** — see #931 for the full rescope. The boundary is enforced by construction + the filesystem, with the app-level gate as fail-fast/observability on top.

**Two flagsets, separated by direction of control:**
- `agent.json` (`agent` domain) — governance (`enabled` kill-switch, `mode`, bounds). **Mounted `:ro`** into the agent container → the kernel refuses agent writes regardless of any app/LLM bug. The agent reads it, never writes it.
- `game.json` (`game` domain) — **writable** by the agent (single flagSetId so game services need no second OpenFeature client). The agent does NOT read game flags into its own decisions.

[...]

PR #953, opened 2026-06-12
feat(agent): shadow-scoped game-flag experiment writer + invariant gate (#932)

aepfli:
## What

Foundation of the M7 self-improvement track — the new `services/agent/experiment/` package lets the agent run **shadow-scoped game-flag experiments**. This is the agreed rescope (not "remote coding agent edits source"): **NO LLM, NO remote coding agent, NO game-service wiring, NO promotion** — just the **Writer + Gate**, fully tested in isolation.

Part of #931/#932 (M7-4/M7-5).

## Design (agreed with maintainer 2026-06-12 — supersedes the original issue bodies)

- **Two flagsets, separated by direction of control.** `game.json` (`game` flagSetId) is the agent's **write** surface; `agent.json` (`agent` flagSetId) is its `:ro`-mounted governance and a **hardcoded deny** here. Single flagSetId → game services need no second OpenFeature client.
- **Shadow-vs-real isolation by construction.** An experiment is written as a flagd targeting rule conditioned on `game_kind != "real"`. Shadow games resolve the experimental variant; real games fall through to the existing `defaultVariant` (and any pre-existing targeting), untouched.
- **THE INVARIANT (the Gate's whole job):** an agent write must **never** change what a `game_kind="real"` evaluation resolves to, for any flag — and the agent may **only** write the game flagset.

### Example (before → after, `invincibility_seconds`, experiment value `6.0`)

```jsonc
// BEFORE
{ "state":"ENABLED",
  "variants": { "2s":2.0, "3s":3.0, "4s":4.0, "5s":5.0, "6s":6.0, "7s":7.0, "8s":8.0 },
  "defaultVariant": "4s" }

// AFTER — new variant + shadow-scoped targeting; defaultVariant + existing variants byte-unchanged
{ "state":"ENABLED",
  "variants": { "2s":2.0, "3s":3.0, "4s":4.0, "5s":5.0, "6s":6.0, "7s":7.0, "8s":8.0,
                "agent_experiment": 6 },
  "defaultVariant": "4s",
  "targeting": { "if": [ { "!=": [ {"var":"game_kind"}, "real" ] }, "agent_experiment", null ] } }
```

For a `real` context the `!=` condition is false → flagd evaluates the else branch (the flag's **pre-existing** targeting, or `null` → `defaultVariant`), byte-identical to before. Targeting shape mirrors the existing `game.json` precedent (the `sensitivity` flag's `{"if":[cond, "fast", null]}`).

## How the Gate enforces the invariant

- **Structural (primary, easy to prove):** the flag must already exist; `defaultVariant` and every game-authored variant are byte-unchanged; the only added variant is the reserved `agent_experiment`; the only targeting delta is a top-level `{"if":[game_kind != "real", agent_experiment, <else>]}`, and the else (real) branch may never select the experimental variant.
- **Belt-and-suspenders eval:** resolve the flag for a synthetic `{game_kind:"real"}` context **before vs after** using the same `diegoholiveira/jsonlogic` engine flagd uses; the resolved variant + value must be identical.
- On reject: writes nothing, returns a typed `*RejectError`, and emits `code_improvement.proposed` with `code_improvement.blocked=true` + reason. On accept: emits the same span with `blocked=false`.
- **Hardcoded denies:** path pinned to the game flagset (`GAME_FLAG_PATH` / `DefaultGamePath`, never an arbitrary path); a Writer mispointed at `agent.json`, or a `FlagKey` smuggling a path, is rejected (belt to the `:ro` mount).

## Writer

Mirrors `actions/writer.go` (#730): order-preserving read-modify-write **in place** under a mutex (no temp+rename → no EBUSY on the flagd bind mount), so untouched flags round-trip byte-for-byte. The Gate is the **single sanctioned entry point** (`Review` applies the write on accept), so no caller can skip validation.

## Package layout

```
services/agent/experiment/
  proposal.go    Proposal{FlagKey, ExperimentalValue, Rationale}
  writer.go      Writer — game.json-only, in-place atomic RMW, shadow targeting
  gate.go        Gate — invariant checks + typed Reason + span; sole entry point
  targeting.go   shadow-targeting builder + shared consts (game_kind/real/variant)
  resolve.go     flagd-style flag resolution (jsonlogic) for the before/after eval
  document.go    order-preserving JSON model (mirrors actions/document.go)
  telemetry.go   code_improvement.* span + attribute consts
  experiment_test.go, testdata/game.json
```

## Tests

`go test -race ./...` green. Coverage: valid shadow proposal (real unchanged, shadow resolves experiment, default/variants byte-stable, span asserted); pre-existing targeting preserved for real; blocked unknown flag / non-game path / path-in-flagkey (file unchanged, blocked span); structural detectors (default change, existing-variant mutation, non-shadow targeting, real-leak in else, real-resolution change via eval); idempotent re-experimentation; atomic concurrent (`-race`); unrelated flags byte-identical.

## Out of scope (later issues)

- Narrative → hypothesis **proposal generation** (the LLM/agent side).
- **Game-services `game_kind` eval-context wiring** — the #838 per-game context threads `game_id`; a follow-up must add `game_kind` before shadow experiments take effect at runtime. **This is the prerequisite.**
- Synthetic fitness validation (#935); promotion to real default (#936); a remote coding agent (#931 backend — likely unneeded).

### DIFF
--- a/services/agent/experiment/document.go
+++ b/services/agent/experiment/document.go
@@ -1,0 +1,207 @@
+package experiment
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+// document / flagObj are an order-preserving model of the flagd game.json file.
+// This MIRRORS services/agent/actions/document.go (the #730 interventions
+// writer): encoding/json drops object key order, which would reshuffle the file
+// on every write and defeat the "preserve unrelated flags byte-for-byte"
+// contract the Gate's invariant relies on (an untouched real-resolving flag must
+// round-trip identically). We keep the document root and each flag we mutate as
+// ordered key lists, and store every value we do NOT touch as the exact
+// json.RawMessage we read.
+//
+// It is a separate copy (not a shared package) because the experiment Writer and
+// the actions Writer target different files with different lock scopes; the
+// machinery is small and the duplication keeps each writer self-contained, as
+// the actions/rollout_writer.go precedent does within its own package.
+type document struct {
+	keys   []string
+	values map[string]json.RawMessage
+}
+
+// flagObj is an order-preserving view of a single flag object.
+type flagObj struct {
+	keys   []string
+	values map[string]json.RawMessage
+}
+
+// parseDoc parses the game flag file into an order-preserving document.
+func parseDoc(raw []byte) (*document, error) {
+	keys, vals, err := decodeOrdered(raw)
+	if err != nil {
+		return nil, fmt.Errorf("parse game flag document: %w", err)
+	}
+	return &document{keys: keys, values: vals}, nil
+}
+
+// flag returns an order-preserving view of the named flag, or an error if it is
+// absent. The Gate requires the flag to pre-exist, so absence is a hard error
+// the experiment surface never invents flags.
+func (d *document) flag(name string) (*flagObj, error) {
+	_, vals, err := d.flagsObject()
+	if err != nil {
+		return nil, err
+	}
+	fr, ok := vals[name]
+	if !ok {
+		return nil, fmt.Errorf("flag %q not present in game flag file", name)
+	}
+	fkeys, fvals, err := decodeOrdered(fr)
+	if err != nil {
+		return nil, fmt.Errorf("decode flag %q: %w", name, err)
+	}
+	return &flagObj{keys: fkeys, values: fvals}, nil
+}
+
+// hasFlag reports whether the named flag exists (used by the Gate before any
+// mutation).
+func (d *document) hasFlag(name string) bool {
+	_, vals, err := d.flagsObject()
+	if err != nil {
+		return false
+	}
+	_, ok := vals[name]
+	return ok
+}
+
+// flagsObject decodes the top-level "flags" object in document order.
+func (d *document) flagsObject() ([]string, map[string]json.RawMessage, error) {
+	flagsRaw, ok := d.values["flags"]
+	if !ok {
+		return nil, nil, fmt.Errorf("game flag document has no \"flags\" object")
+	}
+	keys, vals, err := decodeOrdered(flagsRaw)
+	if err != nil {
+		return nil, nil, fmt.Errorf("decode flags object: %w", err)
+	}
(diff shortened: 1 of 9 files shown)


aepfli:
Adversarial sub-agent review found **two blocking issues** — both now fixed (this is the security-critical foundation, so 'mergeable with nits' wasn't acceptable). The reviewer verified empirically with throwaway probe tests and confirmed the gate's CORE logic is sound: it could NOT change a `game_kind="real"` resolution through any in-process proposal, and the jsonlogic eval fails *closed* on unknown operators (not open). The blockers were at the deployment/path boundary:

**B1 — the `:ro` backstop didn't exist.** `docker-compose.yml` mounted the whole flagd dir writable, so `agent.json` was writable and the code comments asserting a kernel backstop were false. Fixed: added a nested `./services/flagd/agent.json:/etc/flagd/agent.json:ro` mount on the agent service — the kernel now refuses agent writes to its own governance file. The design's belt-and-suspenders is now real.

**B2 — symlink escape.** `isGameFlagPath` was a basename compare, so a `game.json` symlink → `agent.json` passed and the `O_TRUNC` write followed it into `agent.json` (reviewer reproduced it). Fixed: `isGameFlagPath` now rejects a symlinked target (Lstat), and `writeInPlace` opens with `O_NOFOLLOW` (kernel backstop against a TOCTOU swap). `TestReview_BlocksGameJSONSymlink` reproduces the exact attack and asserts `agent.json` is untouched.

Also did **N1**: documented that jsonlogic-errors-on-unknown-operator is load-bearing fail-closed safety (a future lib swap returning false would silently weaken the gate), guarded by `TestResolveFlag_UnknownOperatorFailsClosed`.

Non-blocking N2 (dup-key coalescing), N3 (shadow-only type confusion — worth a `ReasonTypeMismatch` when proposals go LLM-generated), N4 (partial-write availability — accepted, matches #730), N5 (substring heuristic, backstopped by eval) are noted as follow-ups; N3 in particular should land before the LLM generates proposals. `go test -race ./...` green.

PR #954, opened 2026-06-12
feat(flags): thread game_kind into the eval context with fail-safe real default (#932)

aepfli:
## What

Threads the `game_kind` attribute into the OpenFeature evaluation context so the agent's shadow-scoped game-flag experiments (#931/#932) actually take effect at runtime. This is the **runtime prerequisite** the experiment writer's docstring (`services/agent/experiment/targeting.go`) calls out — until now real games always resolved `defaultVariant` because `game_kind` was never in the context.

## The safety property: fail-safe-real-default

The experiment writer scopes targeting on `{"!=": [{"var":"game_kind"}, "real"]}` — **everything that is NOT `game_kind="real"` resolves the EXPERIMENTAL value; only an explicit `"real"` is protected.** That means a context *missing* `game_kind` would resolve the experiment, so a real game could silently run one. The fix makes `"real"` the default everywhere, so any game that doesn't explicitly opt into shadow is protected **by construction**:

- **API-level default** (`lib/feature_flags._init_global_context`): the always-available context now carries `game_kind="real"`. An evaluation with no per-game context still sees `"real"`.
- **Per-game override** (`set_game_transaction_context(..., game_kind="real")`): new param defaults to `"real"`; only a shadow session passes `"shadow"`. Contextvars-scoped, so a shadow session's experiments stay scoped to that session.
- **Session wiring** (`GameSession`): maps its session kind (`primary`/`shadow`, #775) to the eval value (`real`/`shadow`) and sets it via a new `set_game_session_kind_context` **before** `GameFactory.create_game` — so a game mode's `__init__`-time calibration reads (thresholds, grace period, per-mode config) also see the correct split. `BaseGameMode._run` later re-sets the full transaction context carrying the same `game_kind`.
- **Menu** (the real, player-facing game, #837) passes `game_kind="real"` explicitly for clarity.

Shared `GAME_KIND_VAR`/`GAME_KIND_REAL`/`GAME_KIND_SHADOW` constants in `lib/feature_flags.py` pin the values to `targeting.go`'s `GameKindReal = "real"`; a test asserts they match the Go contract so they can't drift.

## Gap investigated (and closed)

`__init__`-time calibration reads happen in `GameFactory.create_game` **before** `_run` set the transaction context. Without the session-boundary `set_game_session_kind_context`, a shadow session's init reads would fall back to the API-level `"real"` default and miss its experiments. This was **fail-safe** (a real game is never put at risk — the worst case was a shadow game not getting its experiment on init-frozen flags), but I closed it anyway by setting `game_kind` at the session boundary before `create_game`. No fail-OPEN path was found: `set_game_transaction_context` is the single transaction-context entry point, every game flows through `GameSession`, and the agent service (Go) does not evaluate the Python game flags directly.

## Tests

- `lib`: API-level default is `"real"`; transaction default `"real"` / shadow override `"shadow"`; constants match the Go contract; a fake-provider test models the `"!= real"` targeting and proves real/unlabeled → baseline, shadow → experimental, and that a context missing `game_kind` would resolve the experiment (why `"real"` must be the default).
- `game_coordinator`: `_eval_game_kind` mapping; the session sets `"shadow"`/`"real"` **before** `create_game` (order asserted).
- Full suites green: `lib` 37, `game_coordinator` 943, `menu` 358. `make lint` clean.

## Out of scope (follow-ups)

- The agent-side experiment writer/gate (#931/#932) — already built in `services/agent/experiment/`.
- Full integration test of a *written* experiment reaching a shadow game (needs the docker stack) — follow-up.
- Promotion (#936).

Part of #932.

### DIFF
--- a/lib/feature_flags.py
+++ b/lib/feature_flags.py
@@ -54,6 +54,15 @@ def _report_exception(flag_key: str, exc: Exception) -> None:
     record_flag_fallback(flag_key, type(exc).__name__, str(exc))
 
 
+# Shadow/real split for the agent's game-flag experiments (#931/#932). These
+# MUST stay byte-identical to services/agent/experiment/targeting.go's
+# GameKindReal ("real") — the experiment writer scopes targeting on
+# {"!=": [{"var":"game_kind"}, "real"]}, so anything that is NOT "real" resolves
+# the experimental variant. "real" is therefore the protected, fail-safe value.
+GAME_KIND_VAR = "game_kind"
+GAME_KIND_REAL = "real"
+GAME_KIND_SHADOW = "shadow"
+
 # Track initialized domains to avoid re-initialization
 _initialized_domains: set[str] = set()
 _hooks_registered: bool = False
@@ -92,6 +101,14 @@ def _init_global_context() -> None:
                 "language": "python",
                 "environment": environment,
                 "hostname": hostname,
+                # Fail-safe baseline for the shadow/real split (#932). The agent's
+                # experiment writer scopes on "game_kind != real" (targeting.go),
+                # so a context MISSING game_kind would resolve the experimental
+                # variant — a real game could silently run an experiment. Defaulting
+                # the always-available API-level context to "real" means any game
+                # that does not explicitly opt into shadow is protected by
+                # construction; only a shadow session overrides this per-session.
+                GAME_KIND_VAR: GAME_KIND_REAL,
             }
         )
     )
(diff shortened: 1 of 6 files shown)


aepfli:
Adversarial safety review: **approve, mergeable as-is**, no blockers. This is the runtime half of the self-improvement safety boundary, so the review hunted specifically for a path where a real player game could run an experiment — and found none. Verified against the actual OpenFeature source:

- **Merge semantics can't erase the default:** `api.merge(transaction).merge(client).merge(invocation)` is per-key, so a per-call empty `EvaluationContext()` (the bare read helpers) cannot wipe the API-level `game_kind="real"`. This was the #1 fail-open risk — it holds.
- **No startup window:** every read helper calls `init_flag_domain` → `_init_global_context` (sets the default) before any provider evaluation.
- **Session-kind mapping is total + fail-safe:** `SHADOW if kind == "shadow" else REAL` everywhere — unknown/empty/None → "real". `BaseGameMode.game_kind` defaults `primary`→real, so even a standalone game is protected.
- **Concurrency isolation:** each `GameSession` runs in its own thread with a fresh event loop → contextvar values are thread-local; a shadow session's `game_kind="shadow"` can't bleed into a concurrent real session.
- Constants pinned byte-identical to `targeting.go` by a test; all suites green (lib 37, game_coordinator 943, menu 358).

Together with #953 (write-time gate), the agent-experiment safety boundary is now adversarially verified at both write time AND evaluation time.

**Noted follow-up (non-blocking):** the targeting test models JSONLogic inline rather than exercising real flagd, so the end-to-end 'a written experiment actually reaches a live shadow game (and never a real one) through real flagd' link isn't yet covered by an automated gate — it needs the docker stack (`make test`). Worth a dedicated integration test before the agent generates real proposals.

PR #955, opened 2026-06-13
feat(agent): reject type-mismatched experiment proposals (#932)

aepfli:
## What

Addresses nit **N3** from the #953 (PR) adversarial review of the experiment writer gate.

The experiment `Gate` accepted a `Proposal.ExperimentalValue any` whose JSON type might not match the flag's existing variants — e.g. a string on a numeric game flag like `invincibility_seconds`. The real-resolution invariant is unaffected (the experiment is shadow-scoped, real games are untouched), but a game-side reader expecting a float could crash or misbehave in a **SHADOW** game once the (later, LLM-generated) proposal is mistyped.

## Change

- New typed reason `ReasonTypeMismatch` (mirrors the existing closed reason set in `gate.go`).
- `checkTypeCompatible` runs in the `Review`/`validate` path alongside the structural checks: the experimental value's JSON **kind** (number/string/bool/array/object) must match the flag's EXISTING non-`agent_experiment` variant values. On a mismatch the proposal is REJECTED before any write — no file change, `code_improvement.blocked=true` + reason on the span.
- **int-vs-float is NOT a mismatch**: JSON has one number kind, so `6` on the `4.0`-variant `invincibility_seconds` flag (and the reverse, a float on integer `num_teams`) is accepted. Null/kindless witnesses impose no constraint; the reserved experimental variant is skipped as a type witness.
- Scope kept to `gate.go` + `proposal.go` + tests only (disjoint from the parallel M7-7 PR).

## Tests

- string on numeric flag → blocked `type_mismatch`, file unchanged, blocked span
- int on float-variant flag, float on int-variant flag → accepted (no false reject)
- bool on bool flag → accepted; number on bool flag → blocked, file unchanged
- unit coverage of `checkTypeCompatible` kind matching + stale-experimental-variant skip

`gofmt -l .` clean, `go vet ./...`, `go build ./...`, `go test -race ./...` all green from `services/agent/`.

Part of #932.

### DIFF
--- a/services/agent/experiment/experiment_test.go
+++ b/services/agent/experiment/experiment_test.go
@@ -347,6 +347,103 @@ func TestCheckRealResolution_DetectsChange(t *testing.T) {
 	}
 }
 
+// ---- Type-compatibility guard (#953 N3) ------------------------------------
+
+// A string on a numeric flag must be REJECTED with type_mismatch before any
+// write — a shadow game's reader of invincibility_seconds expects a float and
+// would crash/misbehave on a string. File unchanged, blocked span emitted.
+func TestReview_BlocksStringOnNumericFlag(t *testing.T) {
+	g, path, rec := gateWithRecorder(t)
+	before := mustRead(t, path)
+
+	err := g.Review(context.Background(), Proposal{
+		FlagKey:           "invincibility_seconds",
+		ExperimentalValue: "six",
+		Rationale:         "mistyped value an LLM might emit",
+	})
+	assertBlocked(t, err, ReasonTypeMismatch)
+	assertUnchanged(t, path, before)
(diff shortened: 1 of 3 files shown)


aepfli:
Reviewed (focused): the type-compat logic is correct — `jsonKind` maps ALL JSON numerics to a single `number` kind so int-vs-float (6 vs 4.0) never false-rejects, null/kindless experimental values and the reserved `agent_experiment` variant are skipped as non-constraints, and a kind mismatch rejects with `ReasonTypeMismatch` before any write (no file change, blocked span). Sits after the unknown-flag check, before the structural write-sim. Tests cover string-on-numeric (blocked), int/float interchange (accepted), bool (accepted/number-blocked), and the stale-variant skip. `go test -race ./...` green post-rebase. Ready to merge.

Issue #778, opened 2026-06-05
Agent: shadow-game runner (start mock-only games from the Go agent)


aepfli:
Part of #774 (shadow games). Phase 4 of 5. **Depends on #730** — reuse the intervention API's transport/client plumbing rather than building a parallel gRPC client stack.

## Goal

`services/agent/gamerunner/`: a `ShadowGameRunner` that launches and drives mock-only games:

1. `MockControllerService.AddControllers(count=N, reserved=true, tag="agent:<id>")` → serials
2. Build `StartGameConfig` with those serials as players
3. `GameCoordinator.StreamGameEvents(StreamEventsRequest{start_config})` (headless start — no menu)
4. Capture `game_id` from the event stream (proto phase adds `GameEvent.game_id`)
5. Drive the game via `SimulateMovement`/`SimulateDeath`/`SimulateButton`
6. On completion/error: `ForceEndGame{game_id}`, remove reserved controllers

## Go module wiring (confirm first)

Go protos generate with go_package `github.com/joustmania/connect-proxy/gen/...`; the agent module is `github.com/joustmania/agent` with no proto dependency today. Either import the connect-proxy gen module (`replace ../connect-proxy` if it has its own go.mod) or add an agent-local buf output. Python services serve plain gRPC, so use the grpc-go stubs (not Connect HTTP).

## Cleanup robustness

Reserved controllers must not orphan on agent crash — the `tag` field enables a sweep of stale reserved controllers.

## Verification

Go unit test against fake gRPC servers (reserve → start → capture game_id → force-end → cleanup). E2E against the compose stack: shadow game visible via `ListGames`, menu lobby stays empty.

aepfli:
Done in PR #833: ShadowGameRunner in services/agent/gamerunner (the agent's first gRPC clients — #730 chose flagd writes, so nothing to reuse), self-contained grpc-go stubs via buf.gen.agent.yaml + make protos-agent, env-gated trigger (AGENT_SHADOW_GAME), orphan sweep by tag, bufconn unit tests + live e2e validation of reservation/cleanup. Flagd-nonce trigger documented as follow-up.

Issue #737, opened 2026-06-04
Remediation observable in traces


aepfli:
## What to build

Make the entire remediation story observable in traces. Every infrastructure decision — expand, rollback, or recommend-only — must be auditable in Jaeger, with the same transparency standard as game decisions.

## Span attributes on `agent.infrastructure.decision`

- `rollout.target` — the backend being rolled out
- `fitness.passing` — whether infrastructure fitness passed
- `fitness.violations` — which fitness functions failed
- `remediation.action` — `"expand"` | `"rollback"` | `"recommended_only"`

## Scope

- All infrastructure decisions traced, including the ones that take no action
- The full incident narrative reconstructable from traces: rollout expanded → unstable backend degraded → fitness violated → rollback executed → fitness recovered
- Tied into the same trace pipeline (OTel Collector → Jaeger on the Pi 5) used for game decisions

## Acceptance Criteria

- [ ] Every `agent.infrastructure.decision` span carries `rollout.target`, `fitness.passing`, `fitness.violations`, `remediation.action`
- [ ] Expand, rollback, and recommended-only outcomes all distinguishable in Jaeger
- [ ] A complete rollout→degradation→rollback sequence is reconstructable from the trace timeline
- [ ] Verified in Jaeger end to end

PR #957, opened 2026-06-13
test(integration): experiment reaches shadow games, never real (#932)

aepfli:
## What this proves

The agent-experiment **safety boundary, end-to-end through the REAL flagd**: a *written* game-flag experiment reaches a **SHADOW** game and **NEVER** a **REAL** game.

    agent experiment writer adds an `agent_experiment` variant + a shadow-scoped
    targeting rule to game.json
      -> flagd file-watch reload (live server, flagSetId=game)
      -> a SHADOW game's eval context (game_kind="shadow") resolves the EXPERIMENTAL value
      -> a REAL game's eval context  (game_kind="real")   resolves the BASELINE value

This is **not** an inline JSONLogic model: the Go `experiment.Gate` evaluates jsonlogic in-process; this test proves the **same targeting block resolves identically on the live flagd** that the game services actually read through, under the exact `game_kind` context values the wiring sets.

## The #954 review follow-up it closes

PR #954 (`feat/game-kind-eval-context`, now merged to `main`) wired `game_kind` into the OpenFeature eval context (`lib/feature_flags`, `GameSession`, the menu). Its review flagged the **one link still lacking an automated gate**: the runtime join — that the #954 wiring + the experiment writer's `!= real` targeting rule + real flagd together honour "shadow only, never real". The Go unit tests cover the writer/gate against an inline jsonlogic model; the #954 Python unit tests cover the context plumbing in isolation. **Neither exercises the actual flagd resolution of the written rule under the real context values.** This test does.

It is the agent-experiment sibling of `test_per_player_targeting_resolves_against_live_flagd` (intervention domain, #730) and `test_rollout_plumbing.py` (rollout domain, #737), following the same live-flagd-RPC + snapshot/restore conventions; kept sequential like the #778/#893 global-flagset tests.

## What it asserts (on live flagd, `flagSetId=game`, RPC resolver port 8013)

Reproduces the exact mutation `services/agent/experiment/targeting.go buildShadowTargeting` makes (no Go writer run — the agent isn't in the integration compose, same as the intervention ActionSink), then resolves under each context:

| context | `invincibility_seconds` | `sensitivity` | meaning |
|---|---|---|---|
| `game_kind=shadow` | 8.0 (experiment) | 0 (experiment) | shadow gets the experiment |
| `game_kind=real` | 4.0 (baseline) | medium=2 | **real protected** |
| `game_kind=real` + `game_mode=Werewolf` | 4.0 | fast=3 | pre-existing real targeting preserved verbatim (experiment wraps it as the else-branch) |
| no `game_kind` | 8.0 (experiment) | 0 | documents that the real-by-default fail-safe lives in the **application** context (`lib/feature_flags`), not flagd — would catch a `== shadow` writer mistake |

A companion `test_game_kind_constants_match_source` pins the `game_kind` var/value + `agent_experiment` variant strings to their single source of truth in **both** languages (Go `targeting.go` + Python `feature_flags.py`), so a contract drift fails loudly rather than silently testing a stale value.

`game.json` is snapshotted and restored byte-for-byte by the `game_flags` fixture so the experiment never leaks.

## How far it ran

Ran locally via the integration harness (`USE_PREBUILT_IMAGES=true USE_DEV_MOUNTS=true pytest tests/integration/test_experiment_shadow_isolation.py`): **2 passed in 26.39s** against the live docker-compose flagd. No flag-file leak after the run. `make lint` clean.

> Note: the original PR base `feat/game-kind-eval-context` merged + was deleted before push, so this was rebased `--onto origin/main` and **retargets `main`** (cascade per plan). #954's `game_kind` wiring is in `main` as commit `732ec03a`.

Part of #932.

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -132,6 +132,40 @@ The integration tests cover:
 - ✅ Distributed tracing propagation across services
 - ✅ Multiple games in sequence
 
+## Experiment shadow-isolation (agent experiment safety boundary, #932)
+
+`test_experiment_shadow_isolation.py` covers the ONE agent-experiment safety link
+no other automated gate covered before it (the #954 review follow-up): that a
+*written* game-flag experiment reaches a SHADOW game and NEVER a REAL game,
+end-to-end through the REAL flagd — not an inline JSONLogic model.
+
+**Scope decision.** Like the intervention/rollout live-flagd tests, the agent
+service (`services/agent`, Go) does NOT run in the integration compose, so we do
+not run the Go experiment Writer here. We REPRODUCE its exact mutation
+(`services/agent/experiment/targeting.go` `buildShadowTargeting`: add the reserved
+`agent_experiment` variant + `{"if":[{"!=":[{"var":"game_kind"},"real"]},
+"agent_experiment", <pre-existing-else-or-null>]}`) into `game.json`, then resolve
+via the flagd RPC resolver (`flagSetId=game`, port 8013) under the EXACT
+`game_kind` eval-context values the #954 wiring sets (`"shadow"` for a shadow
+session, `"real"` for a real game / the fail-safe API-level default). The writer's
+own correctness is the Go unit tests' job (`services/agent/experiment/*_test.go`);
+the #954 context plumbing is covered by Python unit tests — NEITHER exercises the
+actual live-flagd resolution of the written rule under the real context values.
+
+**What this test DOES assert:** a SHADOW context resolves the EXPERIMENTAL value
+(`invincibility_seconds` 8.0, `sensitivity` 0) while a REAL context resolves the
+BASELINE (4.0 / medium), and a REAL game with pre-existing targeting
+(`game_mode=Werewolf`) keeps that targeted value verbatim (the experiment wraps it
+as the else-branch). It also documents WHERE the fail-safe lives: a context with
+no `game_kind` resolves EXPERIMENTAL at raw flagd (rule is `!= real`), so the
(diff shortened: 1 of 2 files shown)


aepfli:
Reviewed (focused — it's a test PR that ran green: `2 passed in 26.39s` against live docker-compose flagd). It genuinely closes the #954-review gap: asserts the full safety matrix through REAL flagd RPC resolution (not an inline JSONLogic model) — `shadow`→experiment (inv 8.0, sens 0), `real`→baseline (inv 4.0, sens 2), `real`+`game_mode=Werewolf`→the pre-existing targeting preserved verbatim (sens 3, proving the experiment correctly wraps prior targeting as the else-branch). The `game_flags` fixture snapshots/restores `game.json` byte-for-byte with `wait_flag_file_restored` (no leak), and `test_game_kind_constants_match_source` pins the var/values to BOTH the Go (`targeting.go`) and Python (`lib/feature_flags`) source so the assertions can't silently test a stale contract.

Valuable nuance it documents: an EMPTY context resolves the experiment at the raw-flagd layer (the rule is `!= real`, and missing ≠ real), so the real-by-default fail-safe lives in the APPLICATION context (`lib/feature_flags`, verified in #954), not flagd — and the test notes this would differ under a `== shadow` rule. That's the same coherent design story as #953 (writer) + #954 (app default). The flagd reload-lag was correctly handled with the harness-wide `FLAGD_RESOLVE_TIMEOUT_SECONDS` + `poll_until` self-heal (a deadline-tuning fix, not a logic bug). `make lint` clean. Ready to merge.

PR #960, opened 2026-06-13
feat(agent): LLM-driven shadow experiment proposer (#931)

aepfli:
## What

M7-4 (#931), on the **agreed rescoped design** (flag experiments, NOT source edits — see the "Refined design 2026-06-12" comments on #931/#932 and `docs/research/m7-4-testing-strategy.md`, PR #958).

A thin **`Proposer`** that turns the agent's rolling game narrative (#929) + fitness signals into a structured `{flag, value, rationale}` experiment and submits it through the existing `experiment.Gate.Review` — **shadow-scoped, safe by construction (#932)**. A full "remote coding agent" is overkill for flag deltas; this reuses the `Backend.Infer` + defensive-decode seam (#739).

```
narrative + fitness  --build prompt-->  Backend.Infer  --decodeProposal-->
    Proposal  --Gate.Review-->  shadow-scoped game.json write (or blocked span)
```

## How it goes through the Gate (shadow-only)

The Proposer **never writes `game.json` directly** — every proposal goes through `Gate.Review`, which enforces THE INVARIANT (a write can never change what a `game_kind="real"` game resolves) + the type guard (#955). So even a hallucinated/malicious model reply can, at worst, write a well-typed **shadow-only** variant; it can never affect real players.

- **Defensive decode** (`proposer_decode.go`, mirrors `llm.Decode`, pure + table-tested): unparseable / incomplete / out-of-vocab => **NO proposal, never fabricated** (#739 discipline). Out-of-vocab flags are rejected before the Gate (kept off the blocked span); the Gate's `ReasonUnknownFlag` is defense-in-depth.
- **Cadence**: `code_improvement.min_interval_seconds` (read live) — proposing is an expensive LLM call, so it fires at most occasionally, not every cycle (mirrors the #847 `llm.min_decision_interval_seconds` floor). Wired on game-end.
- **Engine**: `code_improvement.engine` (read live, flag-selectable per #931 AC) selects the backend; recorded on the span.
- **Span**: `code_improvement.propose_attempt` carries engine + `code_improvement.proposed_diff_summary` (the structured experiment, since the rescope has no source diff) + outcome; the Gate's `code_improvement.proposed` blocked/reason span nests under it.

## What degrades without a real backend

The production `proposeBackend.Infer` is a **sentinel stub** (no real Ollama/cloud transport until #738/#742), exactly like `decision.endpointBackend`. A backend error => the Proposer records `no_backend` and produces **NO proposal, NO error** — the agent simply does not propose. The whole path is wired and inert until a real backend lands.

## File ownership

NEW files only in `experiment/` (`proposer.go`, `proposer_decode.go`, `proposer_telemetry.go` + tests). Added the two flags to `flags/flags.go` + `services/flagd/agent.json` and wired the Proposer in `main.go`. **Did NOT touch** `experiment/gate.go`, `proposal.go`, or `validate.go` (owned/merged), nor the promotion/GitHub side (#936).

## Out of scope (separate issues)

- Validation — #935 (merged; the Validator consumes a proposal separately).
- Promotion to real / GitHub — #936 (parallel PR).
- Real Ollama/cloud inference transport — #738/#742.

Strictly: narrative → propose → submit-to-Gate (shadow-only). It can never affect real games (the Gate guarantees it).

## Tests (in-process fakes, no network)

`inferBackend`-style fake + `tracetest` recorder + temp `game.json` (`GAME_FLAG_PATH`):
- valid reply → Proposal passes `Gate.Review`, writes the shadow experiment.
- undecodable / unknown-flag reply → NO proposal, NO write, recorded.
- known flag + mistyped value → Gate blocks it (`ReasonTypeMismatch` blocked span), no write.
- no backend → degrades to nothing.
- cadence → two cycles within the interval => exactly one proposal.
- pure decoder is exhaustively table-tested (empty / not-JSON / missing-field / null / unknown-flag / valid / object+number round-trip).

`go build`, `go vet`, `gofmt -l`, `go test -race ./...` all green from `services/agent/`.

Part of #931 (M7-4).

### DIFF
--- a/services/agent/experiment/proposer.go
+++ b/services/agent/experiment/proposer.go
@@ -1,0 +1,373 @@
+package experiment
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/joustmania/agent/llm"
+)
+
+// proposer.go is M7-4 (#931): the agent turns its rolling game NARRATIVE (#916/
+// #929) + fitness signals into a structured flag-experiment Proposal and submits
+// it through Gate.Review — shadow-scoped, safe by construction (#932).
+//
+// THE RESCOPED DESIGN (#931/#932 "Refined design 2026-06-12", encoded in
+// proposal.go + the testing-strategy doc docs/research/m7-4-testing-strategy.md):
+// the original "remote coding agent edits SOURCE" is superseded. There is NO
+// source diff. The agent emits a structured {flag, value, rationale} over the
+// EXISTING Backend.Infer + defensive-decode seam (#739) — a full coding agent is
+// overkill for flag deltas. The Proposer is a THIN wrapper over that seam:
+//
+//	narrative + fitness  --Build prompt-->  Backend.Infer  --decodeProposal-->
+//	    Proposal  --Gate.Review-->  shadow-scoped game.json write (or blocked span)
+//
+// SAFETY (the whole point of routing through the Gate, never writing game.json
+// directly): the Gate enforces THE INVARIANT — a write can never change what a
+// game_kind="real" game resolves — and the type guard (#955). So even a malicious
+// or hallucinated model reply can, at worst, write a well-typed shadow-only
+// variant; it can NEVER affect real players. The decoder is the first line
+// (unparseable/out-of-vocab => no proposal at all); the Gate is the structural
+// backstop.
+//
+// DEGRADES TO NOTHING WITHOUT A REAL BACKEND: the production Backend
+// (decision.endpointBackend.Infer) is a sentinel stub until #738/#742 wire a real
+// Ollama/cloud transport. A backend error (the stub, a transport failure, a
+// timeout) is recorded and produces NO proposal and NO error to the caller — the
+// agent simply does not propose this cycle, exactly like the #739 llm path
+// degrades to rules. The engine is flag-selectable via agent.code_improvement.engine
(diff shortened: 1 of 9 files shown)


aepfli:
Adversarial sub-agent review: **approve, mergeable as-is**, no blockers. The untrusted-LLM-output path holds — verified empirically: `ProposeOnce` submits ONLY through `Gate.Review` (no Writer/file bypass); `decodeProposal` is strict + total (empty / not-JSON / malformed / missing-field / null-value / out-of-vocab → no proposal, never fabricated; nil vocab fail-closes; an `agent.json`-smuggle reply is rejected as out-of-vocab before the gate); the model's value lands only as the shadow `agent_experiment` variant and cannot influence the real-resolution targeting (JSONLogic before/after byte-identical). Fires on the single `OnGameEnd` hook for real + shadow games, always producing a SHADOW-scoped experiment by gate construction; cadence-guarded (mutex), degrades to a recorded no-op without a backend, no import cycle, file ownership respected. `go test -race ./...` green. (Also confirmed my two #953 blockers — symlink escape + writable mount — are resolved gate-side: `isGameFlagPath` Lstat-rejects symlinks, Writer opens `O_NOFOLLOW`.)

Three non-blocking nits, all LATENT behind the inert stub backend — tracked as follow-ups for the live-wiring phase (when a real Ollama/cloud backend lands, #738/#742), not churned here:
1. `FitnessSignals` is declared + renders a prompt section but never populated in production (fitness is on the decision engine's `LastFitness`, not `GameContext`) — wire a fitness source or trim the section/claim when wiring the backend.
2. `MinInterval<=0` cadence-disable is test-only (`durationFlag` clamps non-positive → 300s default) — safe; doc it as test-only.
3. The game-end proposer goroutine is ctx-bounded but not joined at shutdown (unlike #917's `AwaitInflight`) and has no per-call timeout — add both when wiring real `Infer`.

Issue #959, opened 2026-06-13
Single source of truth for the active flagd flag-file source (replace implicit .ci.json substitution + scattered CI env checks)


aepfli:
## Problem
Which flagd flag files are active in a given environment is decided **implicitly and in scattered places**:
- `docker-compose.ci.yml` silently substitutes the `.ci` variants by mounting them over the base path, e.g. `./services/flagd/game.ci.json:/etc/flagd/game.json` (also system/controller/observability/user). So in CI, flagd serves `game.ci.json`'s CONTENT at the `game.json` path.
- Code/tests that need to know the *active* flag file have no single source to consult, so they fall back to `os.getenv("CI")` branches (hacky, duplicated, and easy to get wrong).

This bit us concretely in #957 (PR): the experiment shadow-isolation integration test mutated the host `services/flagd/game.json`, but CI's flagd serves `game.ci.json` → the mutation was never served → the test (and its restore) timed out. A test passing locally and failing only in CI for an invisible-substitution reason is exactly the smell.

## Proposal
Define the flag-file source **once**, explicitly, and have both flagd AND the tests/services read it:
- A configurable path/folder (e.g. a `FLAGD_FLAG_DIR` / properties entry) set per environment — base compose points at the normal flag dir, the CI override points at the CI flag set — rather than per-file `:ro` overlay substitution.
- Tests/helpers discover the active flag-file location from that single config instead of branching on `os.getenv("CI")`.
- More flexible than a binary CI true/false (e.g. a third profile, a local experiment dir, etc.).

## Scope
- Introduce the path/folder config (compose + a small accessor used by `lib`/tests).
- Migrate the integration-test flag-file fixtures/helpers and any `os.getenv("CI")` flag-path checks to use it.
- Remove the implicit `.ci.json` per-file substitution in favor of an explicit dir/path selection (or keep file overrides but expose the active path).
- Document the convention.

## Origin
Surfaced by #957 (experiment shadow-isolation test) and the broader self-improvement work (#931/#932) — the experiment writer/tests need to know the active game flagset path. Tracked here so #957 ships with a minimal robust fix and this becomes the proper replacement.

aepfli:
## Design + plan (investigation done)

### Blast-radius inventory (cited)

**Compose (the implicit substitution lives here)**
[...]

Issue #801, opened 2026-06-05
CI flagd configs: game.ci.json missing F1-F3 calibration flags


aepfli:
Found by the M2 completeness audit.

`docker-compose.ci.yml` mounts `services/flagd/game.ci.json` over `/etc/flagd/game.json`, and it lacks the #766 F1-F3 flags (`thresholds`, `zombie.thresholds`, `werewolf.thresholds`, `death_grace_period_seconds`, `nonstop.*`, `fight_club.*`, `tournament.*`, `werewolf.werewolf_fraction`, `traitor.count_tiers`, `zombie.*`, `windows`). In CI these resolve FLAG_NOT_FOUND → code-default fallback. Behavior is correct (promotion was behavior-neutral), but integration tests can never catch a flag-key typo or exercise non-default variants for these flags.

## Acceptance
- [ ] `game.ci.json` carries the F1-F3 flags (defaults matching game.json)
- [ ] A guard (test or CI check) that flags drift between game.json and game.ci.json keys, so this cannot silently regress
- [ ] Verify other *.ci.json domains are not missing M2 additions (agent/interventions have no CI override today — confirm and document that intent)

Refs #766.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #962, opened 2026-06-13
feat(flagd): single source of truth for the active flag-file path (#959)

aepfli:
## Single `FLAGD_FLAG_DIR` env for the active flag directory (#959)

Reworked to the simpler design @aepfli asked for: **one directory env var**, no
naming convention, no overlay-presence accessor, no manifest.

`FLAGD_FLAG_DIR` names the active flag directory. flagd, the Go agent writers,
and the integration tests all resolve the active flag file as a PLAIN join
`$FLAGD_FLAG_DIR/<domain>.json`.

### Why simpler AND more correct
If there is exactly one `<domain>.json` per active flag dir, and flagd + writers
+ tests all resolve `$FLAGD_FLAG_DIR/<domain>.json`, then the file a test mutates
IS the file flagd serves — the mismatch that broke #957 (test wrote `game.json`,
CI flagd served `game.ci.json`) is structurally impossible. The directory is the
single source of truth.

### What changed
- **`lib/flagd_paths.py`** — `flag_dir()` + `active_flag_file(domain)` = plain
  join. Dropped `FLAGD_FLAG_SET`, `has_ci_override`, and the base/ci flag-set
  machinery.
- **`services/agent/flagset`** — `FlagDir()` + `ActiveFlagFile()`/`ResolvePath()`
  = plain join. Per-domain `*_FLAG_PATH` overrides stay as the escape hatch; the
  no-env default is byte-identical to before (`/etc/flagd/<domain>.json`), still
  locked by `TestResolvePathProductionDefaultsUnchanged` for
  game/interventions/rollout.
- **`services/flagd/ci/`** — a COMPLETE set of plainly-named flag files. The five
  formerly-overlaid domains (system/controller/observability/game/user) carry the
  old `.ci.json` content under plain names; the three write-target domains
  (agent/interventions/rollout) are REAL copies of base (not symlinks) so a
  CI-time write never leaks into base.
- **`docker-compose.ci.yml`** — the five per-file `.ci.json` overlay mounts are
  replaced by a SINGLE directory mount (`./services/flagd/ci:/etc/flagd`) for
  flagd. The menu service (the other flag writer) mounts the same CI dir; the
  redundant per-file menu mounts in `docker-compose.dev.yml` are removed so
  `test-dev` does not shadow the CI dir.
- **`conftest`** sets `FLAGD_FLAG_DIR` to `services/flagd/ci`, so
  `helpers.active_flag_file()` resolves the exact served file.
- The #957 experiment test drops its `game*.json` glob stopgap for the single
  `active_flag_file("game")`.

### Removed (the overkill)
The `.ci.json` discovery / overlay-presence branches, the flag-set knob
(`FLAGD_FLAG_SET`), and the parallel Python+Go "active flag set" asymmetry
modelling. Kept to `dir + domain + ".json"`.

### Production write paths unchanged
A deployed agent (no env → default dir `/etc/flagd` + `<domain>.json`) writes to
exactly `/etc/flagd/<domain>.json`, byte-identical to before — test-locked for
game/interventions/rollout.

### Verification
- `cargo`/`go test ./...` (agent) green; `gofmt`/`go vet` clean.
- `make lint` (ruff) clean; 367 lib unit tests pass (incl.
  `test_flagd_paths.py` and the existing `test_flagd_ci_drift.py`).
- **Integration suite against the new CI dir + `FLAGD_FLAG_DIR`** (real stack,
  `test-dev`, all PASSED):
  - `test_rollout_plumbing.py` (2)
  - `test_intervention_flow.py` (3, incl. `test_per_player_targeting_resolves_against_live_flagd`)
  - `test_experiment_shadow_isolation.py` (2 — the #957 test
    `test_experiment_reaches_shadow_never_real`)
  - `test_gameid_intervention_isolation.py` (1 — exercises a REAL menu game +
    shadow game, confirming menu↔CI-dir consistency)

  The test writers wrote `services/flagd/ci/<domain>.json` and the live flagd
  (mounting `services/flagd/ci`) served exactly those mutations.

Part of #959.


aepfli:
Sub-agent review: **approve, no blockers**. Verified the load-bearing invariant — production write paths are byte-identical to before for all four call sites (no-env → `ResolvePath` → `/etc/flagd/<domain>.json`, same as the old `Default*Path` consts), so a deployed agent writes exactly where it did; the Python and Go accessors agree (same base/ci rule, case-insensitive, unknown→base, overlay gated on file presence — their *default dirs* intentionally differ: Python repo-relative for host tooling, Go `/etc/flagd` in-container, both documented); the 3 migrated tests resolve to the files flagd actually serves (interventions/agent/rollout have no `.ci` overlay); purely additive (overlay mounts untouched → flagd serves identical files). lib 12 + Go flagset/actions/experiment all green.

Applied the highest-value suggestion (follow-up commit): extended `TestResolvePathProductionDefaultsUnchanged` from game-only to also lock **interventions** and **rollout** — all three agent write targets are now guarded against FlagDir/const drift (the dangerous vector). The cross-language Python/Go parity test (#954-style shared-fixtures pin) is noted as a worthwhile follow-up; the docstring-scope nit is left as-is.

Mergeable as Increment 1. The Increment-2 decision (finish the knob end-to-end: keep-overlay-and-expose vs folder-reorg) is posted on #959 for the maintainer.

aepfli:
this feels complicated, couldnt i just have provided an envvar via docker compose to set it to a different directory, instead of naming separation, i could have a directory, or just a different docker compose mount? - this feels like overkill for a simple functionality. one envvar containing the directory, which is used by flagd and the logic to manipulate the flag files. or am i worng here?

aepfli:
Follow-up cleanup pushed (commit `a434c035`):

- **Removed the 5 orphaned top-level `.ci.json`** (`controller/game/observability/system/user.ci.json`). They were duplicating the CI flag content that already lives in `services/flagd/ci/` — the rework had left them in place. `ci/` is now the single CI flag source.
- **Drift guard now covers the `ci/` dir** (`lib/tests/test_flagd_ci_drift.py`, #801). It asserts every `ci/<domain>.json` defines the **same flag keys** as base `<domain>.json` (defaults/variants may differ) and that no `ci/` file is orphaned. Stronger than before: all 9 domains in `ci/` are covered, not just the 5 overlaid.
  - This immediately caught real drift: `ci/agent.json` was missing `code_improvement.mode` and `code_improvement.target` (added to base in #936/#961) — exactly the FLAG_NOT_FOUND-in-CI silent-default this guard exists to prevent. Added them.
- **Dropped the dead audio self-mount** in `docker-compose.ci.yml` (`user.ci.json` self-mount) — it pointed at a now-removed file and nothing consumed it (audio reads flags from flagd over gRPC).

Verification: `test_flagd_ci_drift.py` + `test_flagd_paths.py` green (15 passed); `go build/vet/gofmt` clean; merged `docker-compose.yml + override.yml + ci.yml config` exits 0 with flagd/menu mounting `ci/` and audio mounting only assets; integration tests collect cleanly (they already use `FLAGD_FLAG_DIR -> ci/`, unaffected by the removal).

Issue #926, opened 2026-06-10
Flaky: terminal-event timeout in test_parallel_lifecycle (Swapper/Tournament) — distinct from #805


aepfli:
## Context
`test_parallel_mode_lifecycle` intermittently fails with: game did not emit any of `['game_ended','game_force_ended','game_error']` within 40s (session still RUNNING with all players alive).

Observed:
- **Swapper**, fast-batch, CI run 27186348303 / 27186341878 full-build leg (2026-06-09)
- **Tournament**, slow-batch, locally on loaded WSL — reproduces in isolation on base even with the #899/#900 deflake fixes (a43e7571, 852e46c1)

Env-dependent (loaded runner/host), pre-existing — NOT a regression from recent test changes. Distinct from #805 (`game_started` timeout in test_full_game_lifecycle).

## Impact
Beyond red CI: this flake blocks the image-publish jobs (`needs: integration-test`) and caused the 2026-06-09 stale-image cascade on main (see companion CI issue).

## Needed
Root-cause why headless mock games under load fail to converge to a terminal event (mock death-hold vs grace vs EMA timing? event-bus backpressure?) or bound it (per-mode timeout scaling, retry-once for terminal-event waits).

Refs: #899, #900, #826, #805.

aepfli:
Triage decision (2026-06-13): **root-cause first**, do not just bound it. Investigate why headless mock games under load fail to converge to a terminal event — mock death-hold vs grace vs EMA timing, or event-bus backpressure — before adding per-mode timeout scaling / retry-once. Sequenced after the #925 cascade guard + the #805 fix land (those stop the bleeding); this one gets the real fix.

aepfli:
Done — PR #972 merged (2026-06-13). Latch SimulateDeath by a 2-frame DELIVERED-frame budget (not wall-clock), with a ~9.95g death input sized to cross every sensitivity/tempo threshold within 2 frames while exhausting at the kill (so no #757 re-kill); idle controllers clear via a 30s fallback. Fixes the parallel-lifecycle terminal-event timeout under load. Verified across the full SLOW_MAX/FAST_MAX table + a resting-primed test model; rebased onto the post-#974 main so the #969 freshness guard passes.

Issue #925, opened 2026-06-10
CI: decouple service-image publish from flaky integration tests (stale-image cascade)


aepfli:
## Context
2026-06-09 incident: #913 (audio soxr swap) merged, but its CI run failed on a known integration flake → `Build audio` publish job (`needs: [integration-test]`, ci.yml ~419/480) was skipped → no soxr `:latest` was ever published. The next test-only commit took the fast path (`USE_PREBUILT_IMAGES=true USE_DEV_MOUNTS=true`), which mounted soxr-importing source into the pre-soxr image → audio unhealthy → **all 18 integration tests errored on main**, and re-running the red HEAD could never fix it (test-only commits always take the fast path).

Any flake on a deps-changing commit silently leaves `:latest` stale and breaks every subsequent fast-path run.

## Options (pick one or combine)
1. Retry/quarantine known-flaky integration tests so publish isn't blocked by them
2. Publish images on successful **build** + smoke check, decoupled from the full integration gate
3. Fast-path freshness guard: compare pulled image's source revision/deps fingerprint against checkout; fail loudly with a clear message instead of mounting incompatible source

Option 3 is the cheapest and turns a confusing cascade into a one-line diagnosis even if 1/2 are deferred.

Refs: #902/#906 (pipeline restructure), #926 (the triggering flake), incident analysis 2026-06-09.

aepfli:
Done — PR #969 merged (2026-06-13). stale-image cascade guard — per-image deps-fingerprint OCI label + fast-path freshness guard + generic --reruns 1 safety net.

Issue #805, opened 2026-06-05
Flaky: test_full_game_lifecycle[JoustFFA] — game_started timeout (distinct from #757)


aepfli:
## Symptom

`tests/integration/test_full_game_lifecycle.py::test_full_game_lifecycle[JoustFFA]` fails with:

```
TimeoutError: Game did not emit 'game_started' within 25.0 seconds
```

Three occurrences on 2026-06-05, ~12:00–12:10 UTC:
- main run 27013587431 (#787 merge — game-coordinator flag changes only)
- main run 27013764297 (#788 merge — agent-only Go changes, cannot affect game start)
- PR #785 run 27013724627

## Why this is a new flake, not #757

#757 was the death-detection/mode-roulette flake (mock death-hold vs grace vs EMA memory), fixed in #767. This one fails earlier in the lifecycle — the game never **starts**. Two of the three failing runs contain changes that cannot influence the start path, so the cause is on main.

## Suspect

#784 (`reserved mock controllers hidden from button-stream consumers`, merged this morning) changed the `connected_serials` roster / `visible_serials()` semantics that the menu uses for ready-state. A race or off-by-one in the visible-roster handling could intermittently prevent the menu from seeing all controllers ready → no game start. First failure observed shortly after #784 merged (its own PR run also failed).

## Impact

Paints main red intermittently; blocks clean CI on the M3 PR stack (#785/#796+).

aepfli:
Fourth occurrence: run 27014189238 (PR #796, Go-agent-only changes — cannot affect the game start path).

aepfli:
Fifth occurrence: run 27015971439 (PR #813, agent-only Go changes). This is now blocking the M3 merge train roughly every other run.

aepfli:
Done — PR #970 merged (2026-06-13). Root cause CONFIRMED: #784 filtered reserved controllers on the connect/roster path but left the button-event path unfiltered, so the menu auto-registered reserved controllers from button presses while `_reconcile_controllers` (reserved-free roster) pruned them → nondeterministic `all_ready()` mismatch → intermittent `game_started` timeout. Fix: `ButtonDetector` suppresses button events for reserved controllers (reading the same `tracked_controllers` RESERVED source as the connect path — no drift); integration helpers filter reserved serials; regression test `TestReservedControllerSuppression` added.

PR #969, opened 2026-06-13
ci: decouple image publish from flaky integration tests (stale-image guard) (#925)

aepfli:
Part of #925.

## Problem
Image-publish jobs use `needs: [integration-test]`, so a flake on a deps-changing commit skips the publish → `:latest` goes stale. The next test-only commit takes the fast path (`USE_PREBUILT_IMAGES=true USE_DEV_MOUNTS=true`), mounts current source into the stale image, and **all 18 integration tests error on main** with no obvious cause — re-running HEAD can never fix it (the 2026-06-09 soxr cascade).

This implements **Option 3 (primary) + Option 1 (safety net)** from the issue. Option 2 (big restructure) is intentionally not done.

## Option 3 — fast-path freshness guard (keystone)
Turns the silent cascade into a one-line failure *before* any test runs.

- `scripts/ci/deps-fingerprint.sh` — stable 16-char hash of the files that determine what each service image bakes: root/`proto`/`lib`/per-service `pyproject.toml` + every `services/*/Dockerfile`. Deterministic (content + filename, sorted).
- `ci.yml` computes the fingerprint once in the `changes` job (`deps_fingerprint` output) and passes it to `docker-build` as build-arg `DEPS_FINGERPRINT`. Each of the 5 Python service Dockerfiles bakes it as the OCI label `org.joustmania.deps-fingerprint` (`ARG DEPS_FINGERPRINT=dev` → `LABEL ...`).
- `scripts/ci/check-image-freshness.sh` runs as a guard step before the fast path (same `if` as the fast-path step). It pulls each service image (the fast path pulls them anyway) and reads the label via `docker image inspect`, comparing to the checkout. On mismatch:

  > `pulled audio-service:latest predates current deps fingerprint X≠Y — image stale, fast-path unsafe; needs full build.`

  and exits non-zero. **No-op when fresh.** Missing label (pre-guard / never-published image) is also treated as unsafe.

The guard covers exactly the 5 images the dev overlay source-mounts (controller-manager, game-coordinator, menu, audio, pairing-daemon). Infra images are digest/version-pinned and never source-mounted, so they cannot go stale.

## Option 1 — single-flake tolerance (safety net)
- Added `pytest-rerunfailures` and run the integration suite with `--reruns 1` via a generic `PYTEST_EXTRA` passthrough in `make test-integration`. One flake no longer skips publish.
- **Generic**: reruns ANY failure once, never quarantines a test by name. `#805`/`#926` are handled separately.
- A *persistent* failure still fails (only reruns once); a genuinely stale image fails fast at the guard before any reruns.

## Bootstrapping note
This PR edits all 5 service Dockerfiles, so its own merge takes the slow (full-build) path and republishes all 5 `:latest` images **with** the label. From then on the guard has a labeled baseline to compare against.

## Validation
- `python3` YAML load + duplicate-key check on `ci.yml`: clean.
- `make lint`: All checks passed. `make format-check`: 255 files already formatted.
- `uv lock`: resolves with `pytest-rerunfailures`.
- hadolint on the 5 modified Dockerfiles: only the pre-existing `cgr.dev` DL3026 baseline (unrelated to the added `LABEL`).
- Guard logic tested via stubbed `docker`: fresh→exit 0 (no-op), stale→exit 1 (one-line diagnosis), missing-label/pull-fail→exit 1.
- Verified `--build-arg DEPS_FINGERPRINT` → `LABEL org.joustmania.deps-fingerprint` bakes and is read back with the same key via `docker image inspect`.

## Trade-off the reviewer should know
Published service images are **multi-platform manifest lists**; on an index, `buildx imagetools inspect --format .Image.Config.Labels` is a per-digest map and doesn't resolve to one config. The guard therefore `docker pull`s (single runner-arch) and uses `docker image inspect`, which resolves the concrete per-arch label reliably. The fast path pulls these images anyway, so this only warms the cache.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,9 +58,6 @@ jobs:
       docker_build_matrix: ${{ steps.matrices.outputs.docker_build }}
       docker_extra_matrix: ${{ steps.matrices.outputs.docker_extra }}
       retag_matrix: ${{ steps.matrices.outputs.retag }}
-      # Deps fingerprint of this checkout (#925): baked into every service image
-      # at build time and read back by the fast-path freshness guard.
-      deps_fingerprint: ${{ steps.fingerprint.outputs.value }}
     steps:
       - uses: actions/checkout@v6
 
@@ -219,13 +216,6 @@ jobs:
           maybe_retag "$CH_PYTHON_HID"         python-hid
           maybe_retag "$CH_OTEL_COLLECTOR"     otel-collector
           echo "retag=$(json_array "${retag[@]}")" >> "$GITHUB_OUTPUT"
-
-      # #925: single deps fingerprint for this checkout. docker-build bakes it
-      # into each image as a label; the fast path reads it back and refuses to
-      # mount source into an image with a mismatching (stale) fingerprint.
-      - name: Compute deps fingerprint
-        id: fingerprint
-        run: echo "value=$(bash scripts/ci/deps-fingerprint.sh)" >> "$GITHUB_OUTPUT"
 
   # Build and test shared packages (proto, lib)
   lib-tests:
@@ -603,8 +593,8 @@ jobs:
       # image (fingerprint matches). Only runs on the fast path.
       - name: Guard against stale prebuilt images
         if: needs.changes.outputs.integration_deps != 'true' && needs.changes.outputs.force_full != 'true'
+        # IMAGE_PREFIX is already in env scope from the "Set image prefix" step.
         env:
-          IMAGE_PREFIX: ${{ env.IMAGE_PREFIX }}
           IMAGE_TAG: latest
         run: bash scripts/ci/check-image-freshness.sh
 
@@ -663,6 +653,15 @@ jobs:
           IMAGE_NAME="${SERVICE//_/-}-service"
           echo "name=${IMAGE_NAME}" >> "$GITHUB_OUTPUT"
 
+      # #925: per-service deps fingerprint, scoped to exactly the deps inputs
+      # that rebuild THIS image (shared workspace/builder deps + this service's
+      # own pyproject/Dockerfile). Baked into the image as a label below; the
+      # fast path recomputes the same per-service value and refuses to mount
+      # source into an image whose fingerprint is stale.
+      - name: Compute deps fingerprint
+        id: fingerprint
+        run: echo "value=$(bash scripts/ci/deps-fingerprint.sh '${{ matrix.service }}')" >> "$GITHUB_OUTPUT"
+
       - name: Build and push service
         uses: docker/build-push-action@v7
         with:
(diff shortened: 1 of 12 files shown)


aepfli:
## Per-image fingerprints — BLOCKER fixed (`a031b060`)

The guard previously baked **one shared** fingerprint hashing every `services/*/{Dockerfile,pyproject.toml}`, but each of the 5 source-mounted images only rebuilds `:latest` when ITS OWN path filter fires (`shared_python` anchor OR `services/<svc>/**`). So a deps change to a non-guarded service (agent, connect-proxy, dashboard, rust-hid, python_hid) — or to just one of the 5's own `pyproject.toml` — bumped the shared fingerprint without rebuilding the others. After merge those images carried the OLD fingerprint while the checkout computed the NEW one → guard FALSE-fails and wedges the fast path on main. The exact failure this guard exists to prevent.

### Fix — fingerprint is now per-service
`deps-fingerprint.sh <service>` hashes ONLY that service's deps-determining inputs:
- shared baked deps: root `pyproject.toml`, `proto/pyproject.toml`, `lib/pyproject.toml`, `images/builder/**` (Dockerfile + `requirements-common.txt`)
- the service's own `services/<svc>/pyproject.toml` + `services/<svc>/Dockerfile`

This mirrors the per-service rebuild condition (`shared_python` OR `services/<svc>/**`) restricted to deps files. Application `.py` source is excluded (the fast path mounts it fresh).

**End-to-end wiring:** each `docker-build` matrix entry computes its own value (`deps-fingerprint.sh '${{ matrix.service }}'`) → `DEPS_FINGERPRINT` build-arg → Dockerfile `ARG DEPS_FINGERPRINT` → `LABEL org.joustmania.deps-fingerprint` → `check-image-freshness.sh` recomputes the same per-service value and compares per image. The shared `changes.deps_fingerprint` output/step is removed.

### Scoping tests (all pass)
| change | expected | result |
|---|---|---|
| `services/menu/pyproject.toml` | only menu | only `menu` changed |
| `services/audio/Dockerfile` | only audio | only `audio` changed |
| `services/agent/Dockerfile` (non-guarded) | none of 5 | none changed |
| `services/python_hid/pyproject.toml` (own image) | none of 5 | none changed |
| root `pyproject.toml` | all 5 | all 5 changed |
| `proto/pyproject.toml` / `lib/pyproject.toml` | all 5 | all 5 changed |
| `images/builder/Dockerfile` | all 5 | all 5 changed |

Deterministic (same input → same hash on repeated runs); working tree restored to baseline after the test.

### Review nits in the same pass
- **Pull-failure vs missing-label:** `check-image-freshness.sh` now splits `pull_image` from `read_label`. A `docker pull` failure emits a NETWORK/REGISTRY error (retry / check auth — a different operator action) instead of being silently reported as staleness.
- **`--reruns 1` caveat (Makefile):** added a comment noting the session-scoped `docker_compose` fixture with process-global state (rate limiter, `interventions_allowed`) means a green rerun could mask a test-isolation bug — the guard, not the rerun, is the real fix.
- **Redundant `env:` at guard step (ci.yml):** dropped the re-declared `IMAGE_PREFIX` (already in `env` scope from "Set image prefix").

### Validation
- `yaml.safe_load` parse + duplicate-key scan: clean
- `make lint`: All checks passed
- `bash -n` both scripts: clean (`shellcheck` unavailable in this sandbox — no network for install)

PR #972, opened 2026-06-13
fix: converge mock deaths by delivered-frame latch, not wall-clock (#926)

aepfli:
## Root cause (CONFIRMED — mechanism + code path)

`test_parallel_mode_lifecycle` intermittently timed out waiting for a terminal event with the session **still RUNNING and all players alive** (Swapper fast-batch on CI, Tournament slow-batch on loaded WSL). "All alive + RUNNING" means deaths never accumulated to the win condition — the game never converged. The reason: **mock deaths never reached the game under load.**

The death path is EMA-threshold-based:

1. `SimulateDeath` set death-level acceleration (~7.07g) and held it for a fixed **1.0s of wall-clock** (`death_hold_until`).
2. The controller-manager discovery loop polls the mock at 100Hz into the state cache; the **gameplay send loop** (`StreamGameplayData` → `_build_gameplay_update`) samples that cache at its stream rate and streams it to the game.
3. The game's death detection (`base.py::_process_controller_state`) runs an EMA (weight 4) over the streamed accel and kills when it crosses the per-sensitivity threshold. **A single ~7g frame is enough** to cross.

Under 4-way parallel-game CPU load the controller-manager event loop runs four gameplay send loops plus the discovery loop and starves. If the send loop's sampling period stretches past ~1.0s, the **entire 1.0s death-hold window elapses between two sends** and zero death-carrying frames are streamed. The EMA never moves, the death never registers — and re-firing `SimulateDeath` doesn't help, because each fresh 1.0s burst can be skipped the same way.

**Why Swapper & Tournament specifically:** both require *repeated* kills to terminate — Swapper must converge every player onto one team, Tournament must drive the whole single-elimination bracket — so one missed death stalls the whole sequence. A one-shot-elimination mode is far likelier to land its single kill before the budget expires, which is why those modes don't surface the flake.

This is mechanism (a)/(b) from triage (mock death-hold timing × send-loop starvation), not (c) grace/shield or (d) a mode tick.

## The fix (load-independent)

Latch the death by a **delivered-frame budget** instead of wall-clock:

- `SimulateDeath` sets `death_frames_remaining = DEATH_DELIVERY_FRAMES` (8).
- The gameplay send path ticks each streamed controller's latch once per frame via `consume_death_frame`, routed through `MultiplexerBackend` to the owning `MockAdapter`.
- `poll()` reports death-accel while the budget is non-zero; it clears once the budget drains.

The death-level acceleration is now **guaranteed to be streamed in exactly 8 frames** — decisively crossing the EMA threshold — no matter how slowly the send loop runs. Starvation only spreads those frames over more wall-clock; it can no longer skip them. The fix needs **no timeout bump** and makes the test pass on the first try (not via #969's `--reruns 1`).

The wall-clock value is kept only as a safety cap for controllers **not** in a game (no stream draining the latch). Because the budget is drained every frame — including during the 2.0s post-respawn grace window — the latch clears well before grace expires, so it cannot trigger the #757 grace-expiry re-kill.

## Validation

- **Unit:** 656 controller-manager tests pass, including new coverage for the frame-budget hold/drain/safety-cap (`test_mock_adapter`), `MultiplexerBackend.consume_death_frame` routing + no-hook fallthrough, and per-frame draining in `_build_gameplay_update`.
- **Integration under load:** `test_parallel_mode_lifecycle[slow-batch]` (Tournament) + `[fast-batch]` (Swapper) passed **10/10 on the first try (no reruns)** under sustained 16-core CPU saturation (load avg ~16) — the exact condition that starves the old wall-clock latch.
- `make lint`: clean.

Part of #926.

### DIFF
--- a/services/controller_manager/mock_control_service.py
+++ b/services/controller_manager/mock_control_service.py
@@ -35,22 +35,43 @@ logger = logging.getLogger(__name__)
 #
 #   Crossing the death threshold. The game smooths streamed acceleration with an
 #   EMA, weight 4 (smoothed = (smoothed*4 + raw)/5), and kills when it crosses
-#   the per-sensitivity death threshold. The mock is only ever exercised by the
-#   integration tests, which start games at sensitivity 2 (MEDIUM): death
-#   threshold 1.8g at slow music, up to 2.8g at fast music. From rest (EMA
-#   ~1.0g) with the ~7.07g death input the EMA climbs:
-#       frame 1 -> 2.21g     frame 2 -> 3.19g
-#   So frame 1 already crosses MEDIUM/slow (1.8g), and frame 2 (3.19g) crosses
-#   MEDIUM at ANY music tempo (<= 2.8g) AND still kills if frame 1 was dropped.
-#   2 is therefore the minimum count that reliably crosses MEDIUM with one
-#   frame of margin.
+#   the per-sensitivity death threshold. The mock can face ANY sensitivity (0..4)
+#   at EITHER music tempo — the integration flagd config targets Werewolf to the
+#   "fast" profile (sensitivity 3, death threshold up to 3.2g) and other modes
+#   reach sensitivity 4 (3.2g slow / 3.5g fast). The highest threshold the mock
+#   can be asked to cross is therefore 3.5g.
 #
-#   #757-safety by construction. The game-side kill fires on the FIRST death
-#   frame it processes; the killed player is immediately marked not-alive and
-#   given a 2.0s respawn grace during which the game RESETS its EMA every frame.
-#   At most (budget - 1) = 1 death frame is streamed AFTER the killing frame,
-#   and the latch then EXHAUSTS. That single trailing frame lands (budget-1)*S
-#   wall-clock after the kill, where S is the send interval. For it to re-prime
+#   The death input must be sized so that 2 DELIVERED death frames cross that
+#   3.5g ceiling from a resting-primed EMA (~1.0g). With budget 2 the EMA after
+#   the second delivered death frame is (16 + 9*raw)/25, so raw >= ~7.95g is
+#   required for frame 2 >= 3.5g. We use ~9.95g (axes 7,5,5) for comfortable
+#   margin. From rest (EMA ~1.0g) the EMA then climbs:
+#       frame 1 -> 2.79g     frame 2 -> 4.22g
+#   Per-sensitivity kill frame (the frame whose EMA first exceeds the threshold),
+#   verified against the full SLOW_MAX/FAST_MAX tables in
+#   services/game_coordinator/games/base.py:
+#       sens | slow_max kill | fast_max kill
+#         0  |  1.3g  frame 1 |  1.6g  frame 1
+#         1  |  1.5g  frame 1 |  1.8g  frame 1
+#         2  |  1.8g  frame 1 |  2.8g  frame 2
+#         3  |  2.5g  frame 1 |  3.2g  frame 2
+#         4  |  3.2g  frame 2 |  3.5g  frame 2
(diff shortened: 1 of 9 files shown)


aepfli:
## Rework: sizing the death-frame budget to escape BOTH #926 and #757

Thanks for the review — the three blockers were all correct. The diagnosis (starved CM send loop + wall-clock-only hold = skipped death) stands; the implementation is reworked. New commit: `5eae2a03`.

### What was wrong (matches the three blockers)
1. **#926 not actually fixed** — `poll()` ANDed the frame budget with the 1.0s wall-clock cap (`death_hold_until = now + 1.0`). Under the targeted ~1.0s send gap the cap fired *before* the frames were delivered, so the death was still skipped. The `hold_until == 0.0` "no cap" branch was only ever reached by unit tests.
2. **Naive no-cap reintroduces #757** — with budget 8 and no cap, death frames stream for ~8 sends, outliving the 2.0s grace under starvation → grace-expiry re-kill. Verified directly: budget 8, no cap, ~1.25s send gap → **3 re-kills**.
3. **Tests proved nothing** — they only exercised `hold_until == 0.0`, the state production never creates.

### The redesign
**Budget = 2 frames** (`DEATH_DELIVERY_FRAMES`), no early wall-clock cap in-game.

- **Why 2 is the minimum-with-margin for the EMA crossing.** Game EMA is weight 4 (`smoothed = (s*4 + raw)/5`); mock death input is ~7.07g; integration games run at sensitivity 2 (MEDIUM): threshold 1.8g slow → 2.8g fast. From rest (~1.0g): **frame 1 → 2.21g** (crosses MEDIUM/slow), **frame 2 → 3.19g** (crosses MEDIUM at any tempo ≤ 2.8g, and still kills if frame 1 was dropped). One frame of margin; that's the minimum that reliably crosses.
- **Why it exhausts at the kill (#757-safe by construction).** The game kills on the **first** delivered death frame, immediately marks the player not-alive, and opens a 2.0s respawn grace during which it resets the EMA every frame. With budget 2 there is **at most one trailing frame** after the kill; it lands `(budget-1)*S = 1*S` after the kill, comfortably inside the 2.0s grace for any realistic starved gap `S < 2.0s`. (Budget 3 already re-kills at `S ≈ 1.05s`; budget 8 re-kills almost immediately — hence 2, not 3+.) The small budget — not a wall-clock cap — is what makes it safe.
- **In-game vs idle.** In-game the budget *alone* clears the latch (a gameplay consumer drains it via `consume_death_frame`), so a starved send loop can never skip the death (#926). The wall-clock value is now a **generous idle fallback** (`DEATH_IDLE_FALLBACK_SECONDS = 30s`) that only clears a latch nothing is draining (a `SimulateDeath` on a controller not in a running game) — set far above any in-game delivery time, so it can never cut a real death short. Distinction = "is the budget draining?".

### New tests (blocker #3)
- `test_death_lands_under_stretched_send_cadence_926` — fires the **real SimulateDeath RPC**, then runs the actual **two-loop** data plane (100Hz discovery poll into a state cache + a stretched send loop draining the budget) plus a faithful game-side EMA+grace model. **Fails on the old code (death skipped, 0 kills); passes on the rework.**
- `test_no_rekill_after_grace_under_starvation_757` — asserts no re-kill after the 2.0s grace under a starved cadence (guards the budget-8 regression).
- `test_death_lands_at_normal_cadence` — 60Hz sanity.
- Routing/fallthrough + per-frame-drain tests kept (sound per review).

### Validation
- `controller_manager`: **659 passed**; `game_coordinator`: **943 passed**; `make lint` clean.
- `test_parallel_mode_lifecycle[slow-batch=Tournament]` + `[fast-batch=Swapper]`: **4/4 first-try passes, no `--reruns`** — one unloaded, three under sustained 16-core CPU load (load avg up to ~20, the exact starvation that triggered the flake).

Issue #983, opened 2026-06-13
Experiment framework: experiment journal — durable intent + append-only progression + rolling summary (bounded LLM context)


aepfli:
Phase 3 (state) of the agent experiment/cohort framework (epic #982). An experiment is a long-lived, evolving entity; this defines HOW its goal + progression + outcome are stored WITHOUT unbounded context growth.

## Three-part experiment record
1. **Immutable intent (the goal / what we tried)**: experiment_id, hypothesis (what + why), flag(s) + candidate value, fitness objective, arms, target N, stop criteria. Written once at creation.
2. **Append-only event log (the history / progression)**: a chronological ledger — game assigned to arm, game concluded (+ its fitness sample), arm aggregate updated, interim verdict, decision (continue/stop/promote), final outcome. Each new game/conclusion APPENDS one entry; history is never rewritten.
3. **Rolling sufficient-statistics summary (the compact present)**: per-arm count + mean + variance (incremental, Welford), current verdict, short natural-language status. Derived from the log; bounded size.

## Bounded LLM context (the key discipline)
Do NOT feed the LLM the full game-by-game log (grows unbounded). Feed it: the immutable intent + the rolling per-arm aggregates + a short recent-events tail (+ running conclusion). Sufficient statistics (count/mean/variance) keep the context ~constant as N grows from 1 to hundreds of games. Mirrors the existing gamewindow/gamesummary compaction (compact view from a stream) at the experiment timescale; consumed by the retro prompt (#844).

## Durability (survive restart)
The journal (intent + append-only log + rolling summary) must persist across agent restarts (experiments run minutes/hours; #831 flagged in-memory state loss on restart). Authoritative state = a durable local store (JSON/SQLite on a mounted volume); key events ALSO emitted to telemetry for the dashboard (#981). This is the agent OWN working memory — NOT a flag file (distinct from agent.json :ro governance and game.json control surfaces in the M7 design).

## Append model
Game concludes -> append {game_id, arm, fitness, ts} + O(1) incremental aggregate update. Checkpoint -> append {verdict-so-far, decision}. Conclude -> append {final verdict, outcome: promoted/rejected/inconclusive, action}. The log is immutable history; the summary is the compact present.

Foundational: S4 (registry/lifecycle) stores experiments here; S5 (aggregation) reads/updates the rolling stats; S6 (promotion) records the outcome. Relates to #844 (retro prompt), #831 (restart semantics).

aepfli:
[...]

Issue #982, opened 2026-06-13
[EPIC] Agent experiment/cohort attribution framework


aepfli:
## Problem
The agent can tune game flags and observe per-game telemetry tagged with `game_kind` (shadow/real) + `game_id`, partitioned per game by the #845 multiplexer/LoopSet. But there is NO notion of an *experiment* as a cohort of games. The shadow/real split supports isolating experiments from real play, but NOT:
- running K experiments concurrently, each over a cohort of N games,
- scoping a flag override to one experiment games only,
- aggregating fitness per arm across the cohort,
- comparing experimental vs control to a statistically meaningful verdict before promotion.

This is the missing piece that makes the agent self-tuning actually MEASURABLE.

## Current state (origin/main, verified)
- `game_id`: on telemetry (metric label) + adopted as the agent SessionID; multiplexer partitions per game.
- `game_kind` (real/shadow): on telemetry as span attr `game.kind` (game_session.py:187) + metric labels (game_session.py:251-261, servicer.py:465-467); ingested by the agent into `GameContext.GameKind` (extract_metrics.go `attrGameKind` / `SetGameKind`); and an eval-context var (lib/feature_flags.py `GAME_KIND_VAR`, real-by-default). The full eval-context -> telemetry -> ingest -> context loop already works end-to-end.
- Experiment writer scopes targeting on `{"!=":[{"var":"game_kind"},"real"]}` (binary).
- Validator / FitnessMeasurer (#956) measures single runs, not cohorts.

## Model
- Experiment: {experiment_id, hypothesis, flag(s) + candidate value, fitness objective, arms, target N per arm, stop/decision criteria, status}.
- Arm: experimental (candidate value) vs control (shadow-isolated baseline). Optionally multiple experimental arms.
- Cohort: the game_ids assigned to an experiment, split across arms.

## Design (extend the proven game_kind dual-path)
1. Identity propagation: experiment_id + arm as eval-context vars + telemetry (span attrs + metric labels) + agent ingest (GameContext fields) — mirroring game_kind.
2. Spawn binding: shadow games started bound to (experiment_id, arm); cohort membership is ground-truth at start (no drift).
3. Experiment-scoped targeting: targeting keyed on experiment_id/arm so K experiments coexist; real-by-default preserved.
4. Registry & lifecycle: agent-owned coordinator defining experiments, allocating shadow capacity to arms, tracking status + stop criteria.
[...]

Issue #981, opened 2026-06-13
Experiment framework: experiments dashboard view (demo)


aepfli:
Phase 6 (demo presentation) of the agent experiment/cohort framework. Visualize experiments. Belongs to M5 Demo Polish.

Scope:
- A dashboard view (extends the Agent tab #792 / fleet #791) showing live experiments: per experiment the arms, sample counts (games per arm), fitness deltas, and the current verdict; consumes the propagated telemetry + the cohort aggregates.

Acceptance: during a multi-experiment shadow run, the dashboard shows each experiment arms + sample counts + delta + verdict, updating live.

Depends on the identity-propagation + aggregation steps.

aepfli:
Implemented by PR #1045 (squash-merged). Live Experiments tab in the dashboard: queries Loki (service_name=agent, via envoy) for the agent's experiment lifecycle logs, folds per-experiment/per-arm aggregates (sample counts, mean fitness, delta, verdict + teardown reason), polls 5s. Parsing + aggregation verified correct in review; polish follow-ups in #1047. Live multi-experiment verification pending a dry-run.

Issue #979, opened 2026-06-13
Experiment framework: cohort fitness aggregation & arm-comparison verdict


aepfli:
Phase 4 of the agent experiment/cohort framework. The dissect-data-per-experiment core.

Scope:
- Group the per-game GameContexts (the #845 multiplexer / LoopSet) by (experiment_id, arm) using the propagated identity.
- Aggregate the fitness/objective signal across each arm ~N games (mean / variance), compare experimental vs control with a sample-size + significance / effect-size threshold -> a verdict (better / worse / inconclusive).
- Extend the M7 Validator / FitnessMeasurer (#956) from single-run to cohort measurement.

Acceptance: given a running experiment with N games per arm, the agent produces a per-arm aggregate + a comparison verdict; an under-powered cohort (too few samples) yields inconclusive, never a false positive.

Depends on the registry/lifecycle step and the propagated identity.

aepfli:
Done — PR #990 merged. CohortVerdict: min-N + Cohen-d effect-size gate on UNBIASED per-arm variance (M2/(N-1)) from journal.Summary; promote/discard/inconclusive (higher-fitness=better, guaranteed by fitness.go). Swappable strategy (Welch later). Recent-real baseline deferred (tracked).

PR #985, opened 2026-06-13
feat(agent): durable experiment journal — intent + event log + rolling summary (#983)

aepfli:
Part of #982 — durable experiment journal.

Phase 3 (state) of the agent experiment/cohort framework. Adds a self-contained `services/agent/experiment/journal/` package implementing the three-part experiment record from design §9 (`docs/design/agent-experiment-cohorts.md`).

## Three-part record (design §9.1)
- **(a) INTENT — immutable** (`intent.json`): experiment_id, hypothesis, flag_key, experimental_value, objective, target_n_per_arm, arms. Written once at `Create`; never rewritten.
- **(b) EVENT LOG — append-only** (`events.jsonl`, one JSON object per line): `game_assigned`, `game_concluded` (+fitness sample), `interim_verdict`, `decision`, `outcome`. The audit trail and the **only** place raw per-game fitness samples are retained. Never rewritten — O_APPEND + fsync per line.
- **(c) ROLLING SUMMARY — rewritten atomically** (`summary.json`): per-arm `count`/`mean`/`variance` via **Welford** (online, O(1) in N) + the current verdict.

## API
- `Create(root, intent)` → writes `intent.json` + initial `summary.json`.
- `AppendEvent(event)` → appends to `events.jsonl`, folds a `game_concluded` fitness into the matching arm's Welford accumulator, atomically rewrites `summary.json`.
- `Load(root, experimentID)` → **rehydrate**: rebuild the rolling summary from intent + events on startup. The in-memory state is a VIEW of the durable journal (the **#831** in-memory-loss fix); the event log is the source of truth (a corrupt `summary.json` cache is ignored).
- `CompactView()` → the bounded `{intent, rolling summary, short recent-events tail}` the LLM/registry consume — **never** the full log, so prompt size stays ~constant as N grows (§9.2).

## Persistence (design §9.3)
Reuses the proven `gamesummary` pattern: atomic temp + fsync + rename for the rewritten files. Env-overridable dir `AGENT_EXPERIMENT_DIR` (default `/var/lib/joustmania/agent/experiments`), one directory per experiment.

## Scope
Storage + state machinery only — does NOT wire into the registry (#978), aggregator (#979), or telemetry (#981); those consume this package later. Journal backend = **JSONL + JSON** per the epic decision (not SQLite).

## Tests
Welford-vs-direct correctness, append-only ordering/immutability, rehydrate restart-survival (incl. ignoring a corrupt summary cache), atomic no-partial writes, O(1) bounded view, concurrent-append safety. `go test ./... -race`, `go vet ./...`, `gofmt -l .` all clean.

### DIFF
--- a/services/agent/experiment/journal/journal.go
+++ b/services/agent/experiment/journal/journal.go
@@ -76,6 +76,12 @@ func newJournal(root string, opts ...Option) *Journal {
 // exists for this experiment_id — the intent is immutable and write-once, so
 // re-creating would risk silently clobbering an in-flight experiment; the caller
 // should Load an existing one instead.
+//
+// Single-writer-per-experiment: a Journal serializes its own writes with a mutex,
+// but that does NOT coordinate across two handles to the SAME experiment dir.
+// Callers MUST hold at most one live Journal per experiment dir at a time (two
+// handles would interleave appends and racily rewrite summary.json). The registry
+// (#978) owns this single-handle invariant.
 func Create(root string, intent Intent, opts ...Option) (*Journal, error) {
 	if intent.ExperimentID == "" {
 		return nil, fmt.Errorf("journal: intent.ExperimentID is required")
@@ -207,7 +213,13 @@ func (j *Journal) writeSummary() error {
 // pre-restart one. This is the #831 in-memory-loss fix: the in-memory state is a
 // VIEW of the durable journal, reconstructed on startup. summary.json on disk is a
 // cache/telemetry view; the log is the source of truth, so Load does NOT trust it
-// and recomputes from events.
+// and recomputes from events. A torn final event line (a partial append left by a
+// crash) is dropped and truncated by the store so Load still reconverges — see
+// store.readEvents; an interior malformed line stays a hard error.
+//
+// Single-writer-per-experiment: like Create, the returned Journal does not
+// coordinate with other handles to the same experiment dir. Callers MUST keep at
+// most one live Journal per experiment dir; the registry (#978) owns this.
 func Load(root string, experimentID string, opts ...Option) (*Journal, error) {
 	j := newJournal(root, opts...)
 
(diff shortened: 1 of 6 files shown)


aepfli:
**Blocker fixed: torn-tail recovery in `Load`** (commit 81a49a30)

`readEvents` treated any malformed JSONL line as a hard error, so a single torn final line bricked `Load`/rehydrate — contradicting the §9.3 durability claim that "the log is the source of truth … Load reconverges". `appendEventLine` does `O_APPEND` → single write → fsync, and POSIX gives no cross-crash atomicity for an append, so a crash mid-write of event N+1 (after event N was durably fsynced **and acknowledged**) can leave a partial last line.

**Recovery semantics now implemented:**

- **Malformed LAST non-empty line → DROP + TRUNCATE.** That event was never acknowledged to the caller (the summary is rewritten only *after* the append returns), so dropping it is correct and the summary — which trails the log — stays consistent. A warning is logged. The torn bytes are truncated back to the end of the last valid line, so the next `AppendEvent` appends cleanly at the next monotonic `Seq` from the last **valid** event. Truncation (not "append after the partial") is the safe choice: appending after a torn partial line would leave it *interior*, which would then hard-error on the **next** `Load`. End state: a recovered journal stays loadable.
- **Malformed INTERIOR (non-last) line → HARD error.** A later durably-committed event followed it, so it cannot be a torn tail; it signals a genuine lost/corrupted write (consistent with the existing Seq-gap reasoning). Not silently skipped.

Detection uses a deferred verdict: when a line fails to unmarshal we stash it and keep scanning — if another non-empty line follows, the stashed one was interior (hard error); if EOF follows, it was the torn tail (drop + truncate).

**Tests (previously uncovered):**
- `TestRehydrate_TornTrailingLineDropped` — writes 3 valid events, appends a truncated final line with no trailing newline; asserts `Load` succeeds, the rehydrated summary equals the valid-only summary, the file is truncated to the valid lines, a subsequent `AppendEvent` continues at the correct next `Seq`, and the journal remains loadable afterward.
- `TestRehydrate_InteriorCorruptionHardErrors` — a malformed non-last line → `Load` returns an error.

**Also (non-blocking, same pass):**
- Doc note on `ArmView`/`CompactView`: the #979 significance test must consume `Summary()`'s raw `M2` (+count) for an unbiased/Welch estimator — `CompactView` exposes only the population `Variance` (M2/count), which biases small-N.
- Doc note on `Create`/`Load`: single-writer-per-experiment; callers must not hold two live handles to the same experiment dir (registry #978 owns this).

**Validation:** `go test ./... -race` green; new torn-tail test passes; existing 17 journal tests still pass (now 19 total). `go vet ./...` clean, `gofmt -l .` empty. `make lint` (ruff, Python-only) unaffected.

PR #996, opened 2026-06-14
docs: human/agent settings ownership & arbitration model (#820)

aepfli:
Part of #820 (epic #814, milestone M9).

Documents the **ratified** ownership/arbitration contract between the human admin and the adaptive agent — the spec that the M9 correctness fixes (#816, #817) and surfaces (#818, #819) implement against.

## Model
- **Human owns pre-game intent** (`game.json`/`user.json`): game mode, baseline settings, start.
- **Agent owns bounded in-game deltas** (`interventions.json`): overrides/factors/shields with defined composition.
- Agent **never persists** into `game.json`.
- A human baseline change **invalidates** active agent overrides for that parameter.

## Contents
- New `docs/OWNERSHIP_MODEL.md` (275 lines): ownership principle, per-flag ownership table (8 admin flags + agent intervention surface), invalidation rule, composition rules, visibility/kill-switch contracts, source map.
- Linked from `docs/ARCHITECTURE.md` (new "Settings Ownership" section), `CLAUDE.md` Key Documentation, and `services/agent/README.md`.

## Load-bearing rules (depended on by #816/#817)
- **Sensitivity (#816):** human baseline authoritative; agent `global_sensitivity_override` is a transient delta; **restore-on-clear targets the live human baseline** (re-read `sensitivity` from `game.json`), not the frozen `configured_sensitivity` snapshot; per-parameter invalidation. Notes `player_sensitivity_factor`/`global_difficulty_factor` as preferred non-mutating levers.
- **Grace/invincibility/shield (#817):** disambiguates three distinct mechanisms (`grace_until`, `invincible_until`, agent `grant_shield`); rule = **effective grace = max(sources), monotonic-extend never-shorten**; human invincibility is the floor, agent shield additive-only; shield revert is a no-op and never revokes human invincibility.

The doc describes the intended end state with explicit "Today" notes where the code hasn't caught up yet.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #820, opened 2026-06-05
Document the human/agent settings arbitration & ownership model


aepfli:
## Context

Part of #814.

The ownership decisions from the epic and its children need a durable home, following the pattern of the shadow-games analysis doc (#774, PR #781).

## Content

- The ownership model: human = pre-game intent (game.json/user.json), agent = bounded in-game deltas (interventions.json), agent never persists into game.json
- Arbitration rules decided in #816 (sensitivity) and #817 (grace/shield), incl. restore-on-clear semantics
- Write-ownership matrix per flag domain (game/user/agent/interventions) and per actor (menu, agent ActionSink, dashboard, files)
- Guardrail gap table: existing enforcement (`interventions.py:802-831`: allow-list, mode matrix, battery guard, weighted rate-limit) gates the agent only; the human path has no policy layer — state explicitly what is and isn't governed
- Concurrent-write mechanics: both writers use in-place truncate+write (flagd inotify), menu has no cross-service lock, agent has a process-local mutex (`services/agent/actions/writer.go:85`); safe today because files are disjoint — call out that this invariant is load-bearing

## Acceptance Criteria

- [ ] Doc lands in docs/ and is linked from the agent README and #814
- [ ] Every overlapping parameter has a written composition/arbitration rule

aepfli:
Delivered in PR #996 (merged, commit 424cb266). Added docs/OWNERSHIP_MODEL.md documenting the ratified human/agent settings arbitration model (human owns pre-game intent, agent owns bounded in-game deltas, agent never persists to game.json, human baseline change invalidates active agent overrides) with the per-flag ownership table, sensitivity (#816) and grace/shield (#817) composition rules, and the visibility (#818)/kill-switch (#819) contracts. Linked from ARCHITECTURE.md, CLAUDE.md, and the agent README.

Issue #816, opened 2026-06-05
Resolve global sensitivity contention between admin baseline and agent override


aepfli:
## Context

Part of #814. This conflict is live today.

- Admin sets `sensitivity` (game.json) in the lobby; the game coordinator freezes it at game start (`services/game_coordinator/games/base.py:237-243`).
- The agent's `global_sensitivity_override` mutates `game.sensitivity` **mid-game** (`services/game_coordinator/difficulty_handlers.py:67-92`).
- On override clear, the handler restores the **game-start** value — not the admin's current lobby setting.
- Asymmetry: an admin's mid-game sensitivity change silently no-ops, while the agent's change takes effect immediately. To a human operator this reads as "admin mode is broken".

## Open questions / candidate arbitration rules

1. **Admin wins on touch**: an admin baseline change clears any active agent override for that parameter (and notifies the agent via a flag/event so its world model stays consistent).
2. **Override as bounded delta**: agent override becomes a delta/clamp around the admin baseline instead of an absolute index, so baseline changes compose instead of fight.
3. **Restore semantics**: on clear, restore to the *current* resolved baseline, not the value captured at game start.

## Acceptance Criteria

- [ ] Arbitration rule decided (documented in the ownership doc, see #814)
- [ ] `difficulty_handlers.py` restore semantics fixed per the rule
- [ ] Admin mid-game sensitivity change either takes effect or gives explicit feedback that it applies next game
- [ ] Unit tests cover: override active + admin change, override clear after admin change

aepfli:
Delivered in PR #1000 (merged, commit c03c4352). Split the conflated game sensitivity into baseline_sensitivity (admin) + sensitivity_override (agent) with an effective read-only property; agent override never persists into game.json, clear restores the current admin baseline (not the stale game-start snapshot), and an admin baseline change invalidates the active override. 16 new tests. Note: the admin *live mid-game* write surface is forward-compat (no production path mutates baseline mid-game yet) — tracked under admin-surface work #815.

Issue #814, opened 2026-06-05
[EPIC] Define settings ownership between human admin and agent


aepfli:
## Context

The menu's admin mode (controller-driven, `services/menu/handlers/admin.py`) predates the agentic control stack (#722–#734). Today **two actors mutate overlapping game parameters with no precedence rule, no mutual visibility, and no on-device human control**:

- **Human admin** cycles 8 game flags (`sensitivity`, `num_teams`, `random_assignment`, `nonstop.time_limit_seconds`, `invincibility_seconds`, `fight_club.min_rounds`, `werewolf.reveal_time_seconds`, `force_all_start`) by flipping `defaultVariant` in `game.json`/`user.json` via `FlagConfigWriter`. All but `force_all_start` are **frozen at game start** (`services/game_coordinator/games/base.py:237-243`) — mid-game admin changes silently no-op.
- **Agent** writes `interventions.json` only (`services/agent/actions/writer.go`), **live-evaluated mid-game**: `music_tempo_override`, `global_sensitivity_override`, `player_sensitivity_factor`, `shield_seconds`, `volume_override`, plus edge-triggered eliminate/revive/audio_cue/controller_effect/end_game.

The collision is not hypothetical — it exists today:

1. `global_sensitivity_override` mutates `game.sensitivity` mid-game (`services/game_coordinator/difficulty_handlers.py:67-92`) — the exact value the admin set in the lobby. On clear it "restores" the game-start value, not the admin's current setting. Asymmetry: the admin's mid-game change no-ops while the agent's works.
2. Agent `shield_seconds` (`services/game_coordinator/lifecycle_handlers.py:48-73`) and admin `invincibility_seconds` both feed player grace periods with no defined composition.
3. The human gets **zero signal** (LED/audio/dashboard) when the agent changes something — agent activity is indistinguishable from "admin mode is broken again".
4. The admin has **no on-device way to constrain the agent**: `interventions_allowed` lives in `agent.json`, which admin mode cannot touch.

Existing guardrails (`interventions.py:802-831`: allow-list, mode-capability matrix, battery guard, weighted rate-limit) all gate the *agent* — nothing governs the *human/agent interaction*.

## Proposed ownership model (to be settled first)

> **Human owns pre-game intent** (game mode, baseline settings, start). **Agent owns bounded in-game deltas** (overrides/factors with defined composition). The agent never persists into `game.json`. A human baseline change invalidates active agent overrides for that parameter.

## Child issues

[...]

PR #1000, opened 2026-06-14
fix(game_coordinator): layer agent sensitivity override on admin baseline (#816)

aepfli:
Part of #816 (epic #814, milestone M9). Implements the ownership model documented in #820 (PR #996).

## Problem
The agent's `global_sensitivity_override` mutated the single `game.sensitivity` field mid-game. On clear it "restored" the **game-start snapshot** (`configured_sensitivity`), not the admin's current baseline — so a late override clear could clobber a newer admin intent. Meanwhile admin mid-game changes silently no-op. Asymmetric and lossy.

## Fix
Split the conflated `sensitivity` field on the game object into an owner-aware, layered model:
- `baseline_sensitivity` — admin source of truth (pre-game lobby intent)
- `sensitivity_override` — the agent's temporary override layer (`None` = none)
- `sensitivity` — now a **read-only property**: `override if override is not None else baseline`. All existing readers (`_compute_effective_thresholds`, the `game_sensitivity` metric, span attrs) are unchanged.

Writers split by owner:
- **Admin** (`game.sensitivity = X` / `set_baseline_sensitivity`) sets the baseline **and invalidates any active agent override** — human baseline change wins (per the ratified model).
- **Agent** (`apply_sensitivity_override` / `clear_sensitivity_override`) never persists into the baseline (game.json untouched); clear returns to the **current** baseline, not the game-start snapshot.
- `configured_sensitivity` retained as a read-only property for game-start telemetry only — no longer the restore target.

## Arbitration choice
Override is an absolute-index layer (0–4) with invalidate-on-admin-touch, rather than an additive delta — because the agent flag's wire value is an absolute index + `-1` sentinel; treating it as a delta would need an out-of-scope flag-schema/agent change. Invalidate-on-touch already yields the symmetric, admin-wins composition. (Noted for the #820 doc.)

## Tests
- New `tests/test_sensitivity_arbitration.py`: **16 tests** (override apply/clear, admin change during active override, invalidation, clear-restores-current-baseline).
- Characterization-first: 9 passed unchanged against old code (simple paths preserved), 7 new-model tests failed on old / pass on new. Pre-existing `test_revert_restores_configured_sensitivity` still passes unmodified.
- Full game_coordinator suite: **969 passed, 0 failed**. `ruff check` clean.

### DIFF
--- a/services/game_coordinator/difficulty_handlers.py
+++ b/services/game_coordinator/difficulty_handlers.py
@@ -10,9 +10,12 @@ intervention flags against the InterventionManager registry built in PR A:
   resumes the natural schedule from the current state via an explicit revert
   handler (#922 — the manager routes reverts to ``_revert_handlers``, not back to
   the apply-handler, so the override must be cleared there).
-- ``global_sensitivity_override`` — N2: live update of ``game.sensitivity``,
-  consumed next frame by ``_compute_effective_thresholds``; reverting to none
-  (-1) restores the sensitivity the game started with.
+- ``global_sensitivity_override`` — N2: a bounded, temporary override LAYER on
+  top of the admin baseline (never persisted into the baseline — #816 / M9
+  ownership model), consumed next frame by ``_compute_effective_thresholds``;
+  reverting to none (-1) clears the layer so the effective sensitivity returns
+  to the admin's CURRENT baseline (a mid-game admin change wins by invalidating
+  the override).
 - ``player_sensitivity_factor``   — N1: per-serial ``targeting_key`` evaluation
   sets each active player's ``sensitivity_factor`` (clamped 0.5-2.0); serials
   with no targeting match get the neutral default (1.0).
@@ -99,10 +102,14 @@ async def handle_music_tempo_override_revert(ctx: InterventionContext) -> None:
 async def handle_global_sensitivity_override(ctx: InterventionContext) -> None:
     """Apply / clear the agent global-sensitivity override on the live game.
 
-    A value >= 0 sets ``game.sensitivity`` to the matching ``Sensitivity`` index;
-    the change is picked up next frame by ``_compute_effective_thresholds``. A
-    value of -1 (none) restores the sensitivity the game was configured with at
-    start (``game.configured_sensitivity``).
+    The override is a bounded, temporary LAYER on top of the admin baseline; it
+    is never persisted into the baseline (the M9 ownership model — #816). A value
+    >= 0 sets the override to the matching ``Sensitivity`` index, picked up next
+    frame by ``_compute_effective_thresholds``. A value of -1 (none) CLEARS the
+    override so the effective sensitivity returns to the admin's CURRENT baseline
+    (``game.baseline_sensitivity``) — not a stale game-start snapshot. If the
+    admin changed the baseline while the override was active, the override was
+    already invalidated, so this clear is a safe no-op against the new baseline.
     """
     game = ctx.game
     if game is None:
@@ -111,17 +118,15 @@ async def handle_global_sensitivity_override(ctx: InterventionContext) -> None:
 
     idx = int(ctx.value)
     if idx < 0:
-        restored = getattr(game, "configured_sensitivity", game.sensitivity)
-        game.sensitivity = restored
-        logger.info(f"global_sensitivity_override: cleared, restored to {restored.name}")
+        game.clear_sensitivity_override()
         return
 
     try:
-        game.sensitivity = Sensitivity(idx)
+        sensitivity = Sensitivity(idx)
     except ValueError:
         logger.warning(f"global_sensitivity_override: invalid sensitivity index {idx}, ignoring")
         return
-    logger.info(f"global_sensitivity_override: live sensitivity -> {game.sensitivity.name}")
+    game.apply_sensitivity_override(sensitivity)
 
 
 async def handle_player_sensitivity_factor(ctx: InterventionContext, manager: InterventionManager) -> None:
--- a/services/game_coordinator/games/base.py
+++ b/services/game_coordinator/games/base.py
@@ -490,13 +490,28 @@ class BaseGameMode(ABC):
         self.running = False
         self.gameplay_stream = None  # Phase 46: Bidirectional stream for feedback commands
 
-        # Settings - sensitivity is now passed via StartGameConfig
-        # Validate and convert to Sensitivity enum
+        # Settings - sensitivity is now passed via StartGameConfig.
+        #
+        # Sensitivity arbitration (#816, M9 ownership model — epic #814):
+        #   - ``baseline_sensitivity`` is the ADMIN's source of truth (pre-game
+        #     lobby intent). The admin may update it mid-game via
+        #     ``set_baseline_sensitivity`` (which invalidates any active agent
+        #     override — a human baseline change wins).
+        #   - ``sensitivity_override`` is the AGENT's bounded in-game override
+        #     LAYER (``None`` = no override). The agent never persists into the
+        #     baseline; clearing the override returns to the CURRENT baseline,
+        #     not a stale game-start snapshot.
+        #   - ``sensitivity`` (property below) is the EFFECTIVE level read by
+        #     ``_compute_effective_thresholds`` and the metrics/spans: the
+        #     override if active, otherwise the baseline.
+        # Validate and convert to Sensitivity enum.
         if 0 <= sensitivity <= 4:
-            self.sensitivity = Sensitivity(sensitivity)
+            self._baseline_sensitivity = Sensitivity(sensitivity)
         else:
             logger.warning(f"Sensitivity {sensitivity} out of range, using MEDIUM")
-            self.sensitivity = Sensitivity.MEDIUM
+            self._baseline_sensitivity = Sensitivity.MEDIUM
+        # No agent override at game start; effective == baseline.
+        self._sensitivity_override: Sensitivity | None = None
 
         # #766 F1: death/warning threshold tables promoted to the ``game.thresholds``
         # object flag. Read ONCE here (init-frozen by design — no live re-evaluation);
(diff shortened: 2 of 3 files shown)


services/game_coordinator/games/base.py
```diff
@@ -1500,6 +1590,48 @@ async def _warn_player(self, serial: str, accel_mag: float, threshold: float):
...
+        only ever EXTEND protection (``grant_shield`` is extend-not-shorten on
+        ``grace_until``); it can neither shorten the admin baseline nor be
+        shortened by it. Symmetrically the admin baseline never silently cancels
+        an active agent shield. ``max`` is the only rule that honors both
+        ownership boundaries at once — overwrite or last-write-wins would let one
+        actor silently void the other.
+
+        Every death/immunity check (the base per-frame gate and the mode-specific
+        ``_kill_player_impl`` / ``_process_controller_state`` checks in Tournament
+        and Fight Club) routes through this helper so the rule lives in exactly
```
aepfli: This docstring claims the mode-specific `_kill_player_impl` / `_process_controller_state` checks in Tournament and Fight Club route through this helper. In THIS PR they do not: `fight_club.py:673`, `fight_club.py:410`, and `tournament.py:666` still check `invincible_until` only. Fight Club fully overrides `_process_controller_state` (no super()), so the base call site rewired below never runs for it. The agent shield (grace_until) is therefore still ignored in those modes here. The actual wiring lives in PR #1002 (#817) — this whole helper belongs there, not in the #816 PR.

services/game_coordinator/games/base.py
```diff
@@ -1500,6 +1590,48 @@ async def _warn_player(self, serial: str, accel_mag: float, threshold: float):
...
+        one place.
+        """
+        # base Player has no invincible_until; only Tournament/Fight Club add it.
+        return max(player.grace_until, getattr(player, "invincible_until", 0.0))
+
+    def is_in_grace(self, player: "Player", now: float | None = None) -> bool:
```
aepfli: `is_in_grace` is dead code in this PR — no call site (verified by grep). It is wired into the mode kill paths only in PR #1002. This is #817 scope leaking into the #816 PR.

services/game_coordinator/games/base.py
```diff
@@ -1405,9 +1493,11 @@ async def _process_controller_state(self, controller_state):
         accel_mag = self._compute_accel_magnitude(accel)
 
         # Check grace period first - no death or warning during grace period
-        # Matches original JoustMania: if time.time() > no_rumble
+        # Matches original JoustMania: if time.time() > no_rumble. Uses the unified
+        # effective-grace rule (#817): max(agent shield grace_until, admin
+        # invincible_until) so the agent shield and admin baseline compose here too.
         current_time = time.time()
-        if current_time < player.grace_until:
+        if current_time < self.effective_grace_until(player):
```
aepfli: This is the only `effective_grace_until` call site this PR actually rewires, but it is on the BASE `_process_controller_state`, which Fight Club fully overrides and Tournament's kill path bypasses. So for the two modes that have `invincible_until`, this change has no effect — the composition is incomplete without the mode-path wiring that PR #1002 adds. Recommend removing all #817 grace code from this #816 PR.

services/game_coordinator/difficulty_handlers.py
```diff
@@ -111,17 +118,15 @@ async def handle_global_sensitivity_override(ctx: InterventionContext) -> None:
 
     idx = int(ctx.value)
     if idx < 0:
-        restored = getattr(game, "configured_sensitivity", game.sensitivity)
-        game.sensitivity = restored
-        logger.info(f"global_sensitivity_override: cleared, restored to {restored.name}")
+        game.clear_sensitivity_override()
```
aepfli, line 121: Clean delegation. The idx<0 -> clear and Sensitivity(idx) apply match the flag schema (variants none=-1, 0..4) and the agent writer's absolute-index setState exactly. Sensitivity half of this PR is correct.

aepfli (commented):
> Note: posting as COMMENT (GitHub blocks REQUEST_CHANGES on a self-authored PR). **Effective verdict: Request changes** — see items 1-3 below.

## Review: #816 sensitivity arbitration

**Overall:** The sensitivity half of this PR is well-designed and correct. The split into `baseline_sensitivity` (admin) / `sensitivity_override` (agent, None=none) / `sensitivity` (read-only effective property) is clean, every reader transparently sees the effective value, and no production writer is broken by the property conversion. All 16 new tests pass, plus the full intervention/difficulty/base suites (345 tests) pass green against the refactor. The absolute-index (0-4, -1 sentinel) handling matches the flag schema and the agent writer end-to-end.

**However, this PR has a serious scope problem that blocks merge as-is:** it bundles a large, INCOMPLETE copy of the #817 grace-composition work that already has its own dedicated PR (#1002), and that copy ships a docstring making a claim that is false in this PR.

### Blocking

1. **Unannounced + incomplete #817 code that collides with PR #1002.** This PR adds `effective_grace_until`, `is_in_grace`, the `grant_shield` rewrite, and the `_process_controller_state` grace-check change — none mentioned in the PR description (which is scoped to #816). PR #1002 ("unify ... grace composition (#817)") adds the SAME base.py primitives AND actually wires `is_in_grace` into the Fight Club / Tournament kill paths plus ships grace tests. These two PRs will conflict on merge of base.py. The #817 changes should be removed from this PR entirely and left to #1002.

2. **The `effective_grace_until` docstring is false in this PR.** It states: *"Every death/immunity check (the base per-frame gate and the mode-specific `_kill_player_impl` / `_process_controller_state` checks in Tournament and Fight Club) routes through this helper."* It does not. `fight_club.py:673`, `fight_club.py:410`, and `tournament.py:666` still check `invincible_until` only and never call the helper. Fight Club fully overrides `_process_controller_state` (no `super()`), so the one base call site this PR rewires never even runs for it. Net effect: in this PR the agent shield is STILL silently ignored in both modes' death paths — the exact latent gap the ownership doc section 4.2 rule 4 says #817 must close, left unclosed.

3. **`is_in_grace` is dead code here.** Defined but called nowhere in this PR (only #1002 wires it in). Either both helper + wiring belong here, or neither does — and per (1) neither does.

### Important

4. **#816 acceptance criterion not fully met.** AC: *"Admin mid-game sensitivity change either takes effect or gives explicit feedback that it applies next game."* There is no production path that calls `set_baseline_sensitivity` / assigns `game.sensitivity` mid-game (verified: only tests do). Sensitivity stays init-frozen, so `_baseline_sensitivity` == game-start value for the whole game. Consequently `clear_sensitivity_override` restoring the "current baseline" is observationally identical to restoring the game-start snapshot today — the invalidation machinery is wired for a future admin live-write path that does not exist. That is reasonable forward-compat, but the admin-mid-game-change AC is not actually delivered. Recommend either implementing the live admin path / "applies next game" feedback, or explicitly noting in the PR + #816 that this AC is deferred.

5. **Design deviation from ownership doc 4.1 rule 2 (acceptable, but document it).** The doc says restore-on-clear should re-read the LIVE `sensitivity` flag from game.json by game_id (fail-safe to `configured_sensitivity`). This PR instead restores the in-memory `_baseline_sensitivity`. The PR body justifies invalidate-on-touch over live-re-read, and the doc's own implementation note allows the minimal version — so this is fine, but the doc's 4.1 rule 2 and the implementation now diverge. Update OWNERSHIP_MODEL.md (or note the chosen variant) so the spec and code agree.

### Nits

- `clear_sensitivity_override` when no override is active is an idempotent no-op (correct). A test for "sentinel clear with no active override" would be a cheap addition to the otherwise-thorough suite.

### Verdict

**Request changes.** The #816 sensitivity work is solid and I'd happily approve it on its own. The blocker is the bundled, incomplete, soon-to-conflict #817 code (items 1-3). Strip the grace/shield changes out and leave them to PR #1002; then this is an easy approve once items 4-5 are addressed (or explicitly deferred).

aepfli:
## Update — addressed review

Thanks for the thorough review. Resolved the blocking items:

**Blocking (1–3): leaked #817 grace code removed.** The `effective_grace_until` / `is_in_grace` helpers, the `grant_shield` rewrite, and the base per-frame grace-gate change were cross-contamination from a parallel worktree mishap — they did **not** belong in this PR. All three grace hunks are reverted; `base.py` here is now **sensitivity-only** (0 grace references in the diff). The complete, correct grace composition lives in #817 / PR #1002 (which the reviewer separately approved, with a full kill-path sweep across all 13 modes).

Also **rebased onto current `main`** now that #820 (PR #996) is merged, so the diff is clean (no phantom doc deletions).

Re-verified after the cleanup: full `game_coordinator` suite **969 passed**, `ruff check` clean.

**Non-blocking items, for the record:**
- **(4) AC "admin mid-game change":** Correct — no production path mutates the baseline mid-game today, so the invalidation/restore-to-current-baseline machinery is forward-compatible rather than observably different from the old snapshot behavior *yet*. The actual reported contention bug (clear restoring a stale game-start snapshot instead of current intent) **is** fixed and the ownership boundary is now structurally enforced. The live admin mid-game write surface is tracked separately (admin-surface work, #815).
- **(5) Doc vs impl (OWNERSHIP_MODEL.md §4.1 rule 2):** Implemented the minimal in-memory baseline + invalidate-on-touch variant, which the doc's own implementation note permits. Will align the doc wording in a small follow-up so spec and code read identically.

aepfli (commented):
**LGTM — would approve** (self-approval blocked because author == reviewer). Confirmation re-review following review 4492632220.

The leaked #817 grace-period code is fully removed. Verified:

- **Diff is SENSITIVITY-ONLY**: `difficulty_handlers.py`, `games/base.py`, `tests/test_sensitivity_arbitration.py` (new). No other files touched.
- **ZERO grace contamination**: no `effective_grace_until` / `is_in_grace` references; the only `grant_shield` hit is the pre-existing `base.py:1591` method definition, untouched by this diff (confirmed via `git diff origin/main...HEAD`). No grace-gate changes.
- **Sensitivity fix intact & correct**: read-only effective `sensitivity` property (override if set, else baseline); admin assignment routes through the setter → `set_baseline_sensitivity`, which invalidates any active agent override; `clear_sensitivity_override` restores the CURRENT baseline (not a stale game-start snapshot); no writer writes through the read-only property to raise — the setter cleanly delegates. `configured_sensitivity` preserved as a telemetry-only property.
- **No new issues from the removal**: no dangling helper refs, no broken imports, no dead code.
- **Rebase correct**: HEAD sits directly on top of the merged #820 ownership-doc commit (424cb266); `origin/main` is an ancestor; no phantom deletion of `docs/OWNERSHIP_MODEL.md` or other already-merged files.
- **Tests**: 40 passed (`test_sensitivity_arbitration.py` + `test_difficulty_interventions.py`). The pre-existing `test_revert_restores_configured_sensitivity` stays green because with no admin baseline change `configured == baseline`, so the two test files are mutually consistent.

Remaining items are the two previously-noted **non-blocking** ones:
1. Admin live mid-game baseline write is forward-compat (the setter/`set_baseline_sensitivity` path exists; no wired admin surface drives it yet).
2. Doc §4.1 wording alignment — follow-up.

Verdict: would **Approve**, merge when ready.

Issue #817, opened 2026-06-05
Unify grace-period composition: admin invincibility_seconds vs agent shield_seconds


aepfli:
## Context

Part of #814.

Two actors grant temporary death-immunity through different paths with no defined composition:

- **Admin**: `invincibility_seconds` (game.json), read at game init, feeds mode-specific grace (Tournament/FightClub `invincible_until`, NonStop spawn protection).
- **Agent**: `shield_seconds` (interventions.json, per-serial targeting), live-evaluated (`services/game_coordinator/lifecycle_handlers.py:48-73`), calls the game's `grant_shield`.

If an agent shield lands while mode-level invincibility is active (or vice versa), behavior is whichever write happens to land last — max vs overwrite vs extend is undefined.

## Proposed direction

Single grace primitive in `BaseGameMode`: `player.grace_until = max(existing, now + granted)`. Both admin-config invincibility and agent shields route through it. Document that shields never *shorten* an active grace period.

## Acceptance Criteria

- [ ] One code path owns `player.grace_until`; both sources route through it
- [ ] Composition rule (suggest `max()`) decided and documented
- [ ] Unit tests: shield during invincibility, invincibility during shield, repeated shields

aepfli:
Delivered in PR #1002 (merged, commit 65c0efb9). Unified grace composition: effective_grace_until = max(grace_until, invincible_until), monotonic extend-never-shorten, via a single is_in_grace()/effective_grace_until() helper on BaseGameMode routed through every kill path. Also fixed a real latent bug — Fight Club's _process_controller_state override bypassed the base grace gate, so agent shields were silently ignored there. 18 new tests; full kill-path sweep across all 13 modes verified in review.

PR #1002, opened 2026-06-14
feat(game_coordinator): unify admin invincibility + agent shield grace composition (#817)

aepfli:
Part of #817 (epic #814, milestone M9). Implements the grace composition specified in #820 (PR #996).

## Problem
Admin `invincibility_seconds` and agent `shield_seconds` both feed player grace with **no defined composition** — and they were checked at different layers on disjoint fields. The base per-frame death gate checked only `grace_until`; Tournament/Fight Club checked only `invincible_until` inside `_kill_player_impl`. **Fight Club overrides `_process_controller_state` and bypassed the base `grace_until` gate entirely**, so agent shields (`grant_shield` → `grace_until`) were **silently ignored in Fight Club**. This is a real bug, not just an undefined rule.

## Composition rule
```
effective_grace_until(player) = max(grace_until, invincible_until)
is_in_grace(player)           = now < effective_grace_until(player)
```
A player is death-immune while **either** window is open — monotonic-extend, never-shorten. The agent owns `grace_until` (extend-only via `grant_shield`) and never writes `invincible_until` (the admin baseline). `max()` is the only rule honoring both ownership boundaries; neither source can silently cancel the other. Single-sourced as a static helper on `BaseGameMode` with the full ownership-model docstring.

## Changes
- `games/base.py` — `effective_grace_until()` + `is_in_grace()`; per-frame death gate and `grant_shield`'s extend guard routed through the helper.
- `games/tournament.py` — `_kill_player_impl` uses `is_in_grace` instead of bare `invincible_until`.
- `games/fight_club.py` — both kill checks use `is_in_grace` (fixes shields being ignored).
- `tests/test_grace_composition.py` — new, 18 tests.

NonStop already reaches the base gate via `super()` (shields worked); its `spawn_protected` is a separate respawn mechanism, left untouched. `invincibility_seconds` maps only to Tournament/Fight Club per `game_factory.py`.

## Behavior change
The only intended behavioral change: **Fight Club now honors agent shields** (covered by `test_agent_shield_protects_fight_club_fighter`). Admin invincibility still blocks death in both modes (characterization tests pass) — no regressions. `grant_shield`'s "already covered" guard now compares against effective grace, so a short shield under a longer admin invincibility is correctly a no-op and never writes a `grace_until` shorter than the baseline.

## Tests
- New suite: **18 passed**.
- Full game_coordinator suite: **971 passed, 0 failed**. `make lint` clean.

### DIFF
--- a/services/game_coordinator/games/base.py
+++ b/services/game_coordinator/games/base.py
@@ -1405,9 +1405,11 @@ class BaseGameMode(ABC):
         accel_mag = self._compute_accel_magnitude(accel)
 
         # Check grace period first - no death or warning during grace period
-        # Matches original JoustMania: if time.time() > no_rumble
+        # Matches original JoustMania: if time.time() > no_rumble. Uses the unified
+        # effective-grace rule (#817): max(agent shield grace_until, admin
+        # invincible_until) so the agent shield and admin baseline compose here too.
         current_time = time.time()
-        if current_time < player.grace_until:
+        if current_time < self.effective_grace_until(player):
             # During grace, reset the EMA so it re-primes from the first
             # post-grace frame. Invincibility must *forget* the death spike:
             # otherwise the EMA still holds it when grace expires and the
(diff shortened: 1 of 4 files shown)


aepfli (commented):
## Review: #817 unify grace/shield composition

**Verdict: Approve (posting as COMMENT since GitHub blocks self-approval).** The sweep is complete, the `max()` rule is applied correctly everywhere it needs to be, the extend-only invariant on `grant_shield` is preserved, and the real latent Fight Club bug (its own `_process_controller_state` gate ignored `grace_until`) is genuinely fixed. 18/18 tests pass locally in the worktree. This matches the ratified contract in OWNERSHIP_MODEL.md section 4.2: effective grace = max(grace_until, invincible_until), monotonic-extend-never-shorten, agent owns grace_until and never writes invincible_until.

### Completeness sweep (the highest-risk part) — VERIFIED COMPLETE

I enumerated every death/elimination code path across all 13 game classes. There are exactly two gating surfaces, and both are now unified:

1. **Motion-death gate** lives in `_process_controller_state`. Only three classes override it:
   - `base.py:1412` -> now `effective_grace_until` (covers ffa, teams, simple_teams, random_teams, traitor, werewolf, zombie, swapper — all inherit it).
   - `fight_club.py:678` -> now `is_in_grace` (was the bug: its override bypassed the base gate and only checked `invincible_until`).
   - `nonstop_joust.py:245` -> gates on `spawn_protected` then calls `super()._process_controller_state`, which now uses `effective_grace_until`. So an agent shield IS honored in NonStop via the base gate. Correctly left otherwise untouched.

2. **Forced-elimination gate** lives inside per-mode `_kill_player_impl` (reached by the `eliminate_player` agent intervention via `lifecycle_handlers.py:112 -> game._kill_player`, which bypasses `_process_controller_state`). Only Fight Club (`fight_club.py:411`) and Tournament (`tournament.py:668`) have such a check, and both now route through `is_in_grace`. The base `_kill_player` template (base.py:1710) intentionally does no grace gating.

No other mode has a kill-path override that checks immunity, so nothing is missed. Confirmed ffa / teams_base / traitor / werewolf / zombie `_kill_player_impl` have no immunity check (and no admin baseline maps to them — see below — so nothing to compose there).

### Priority 5 (factory mapping) — VERIFIED
`game_factory.py:185,191` map `invincibility_seconds` ONLY to Tournament and Fight Club. NonStop (176-180) gets only `time_limit_seconds`. So the only two modes carrying an admin `invincible_until` baseline are exactly the two that now check it through the unified helper.

### Priority 4 (NonStop spawn_protected) — VERIFIED
`spawn_protected` is a boolean respawn mechanism driven by `nonstop.spawn_protection_seconds`, not `invincibility_seconds`. Genuinely separate, correctly untouched. Agent shields still compose for NonStop via the base gate.

### Priority 2 & 3 (max + extend-only no-op) — VERIFIED
`grant_shield` (base.py:1592) guards on `new_grace <= effective_grace_until(player)`, so:
- A shorter shield under a longer admin invincibility is a true no-op: returns True, never writes `grace_until`, so it can neither shorten the admin baseline nor write a `grace_until` below it.
- It also can't shorten an existing longer `grace_until`.
- It writes only `grace_until`, never `invincible_until`.

### Test quality (priority 6) — strong
All required cases covered: both-windows-open (shield longer / invincibility longer), shield-shorter no-op, neither-active, base-gate honoring admin invincibility, and the headline Fight Club shield-now-honored case (both `_process_controller_state` and `_kill_player_impl`). Tournament has both directions. Good characterization tests guard the pre-existing admin-invincibility behavior too.

### Non-blocking observations (not for this PR)

1. **`eliminate_player` vs an agent shield is now asymmetric across modes.** In Fight Club and Tournament an active grace window (incl. an agent shield) blocks an `eliminate_player` intervention; in every other mode eliminate is not grace-gated. This asymmetry is *pre-existing* — the PR only swapped `invincible_until` -> `effective_grace_until` in the two modes that already had a check, so it is not a regression. Worth a follow-up to decide intentionally whether an agent shield should block an agent eliminate, and make it consistent (a single grace check in base `_kill_player` would do it).

2. **`max` semantics absorb (rather than stack) a shorter shield under a longer admin window.** Intended per the ratified rule; just documenting the corner that if admin invincibility expires earlier than the absorbed shield's nominal end, the shield is effectively lost.

Nice work — the `effective_grace_until` docstring is excellent and the call-site comments explain the *why* at each gate.

Issue #1140, opened 2026-06-16
Trace correlation Phase 3 (#1088): stitch disjoint agent trace roots + instrument the retro/experiment lifecycle


aepfli:
Follow-up to #1088. Phases 1 (#1095 game.id attr) and 2 (#1133/#1136 decision→game span Links) are merged, but the agent's trace story is still disjoint in three places. Audit of the live span topology (signal→decision→game→retro):

## Connected today
- **Sync/rules path**: `agent.signal_received` (root) → `agent.decision` → `agent.action` — one clean trace, unbroken parent chain.
- **Decision↔game**: `game.id` attribute (#1095) + span Links (#1133/#1136). Navigable cross-trace reference (not shared trace — by design; coordinator doesn't propagate trace ctx to the agent).
- **Inference→gateway**: W3C traceparent (#1112/#1123) parents the litellm gateway span under the agent's infer-call span.

### A. Async (LLM) decision smeared across 3 roots — HIGH (demo path)
When inference fires async (#917), the evaluate cycle returns `nil` (no decision span in the `signal_received` trace); the result lands later under a NEW root `agent.llm.apply`. One logical decision = three disjoint traces: `signal_received`, `llm.infer.call` (+gateway), `llm.apply` (+decision/action).
- **Fix**: carry the fire-cycle `SpanContext` through the async job and add span Links from `llm.infer.call` + `llm.apply` back to the originating `agent.signal_received` (reuse the #1133 Link primitive). Don't block the loop (#917 design preserved) — Links, not parenting.
- Evidence: `decision.go:735` `if l.fireInferAsync(...) { return nil }`; `async_apply.go:50` `Tracer.Start(root, SpanLLMApply, ...)` where `root=l.rootCtx` (no parent); `async_infer.go:251` `SpanLLMInferCall` started on a parent-less ctx.

### B. Retro orphaned from the game AND from its decisions — HIGH (stated priority)
`agent.llm.retro` is created with `context.Background()` → orphaned root, tied to the game only by `game.id` attribute, with NO Link to the in-game decisions it reflects on.
- **Fix**: Link `agent.llm.retro` → the game trace (reuse `GameContext.GameTraceID` from #1133) and → the decision spans it analyzes.
- Evidence: `retro_capture.go:290` `rc.Tracer.Start(context.Background(), SpanLLMRetro, ...)`.

### C. Experiment/fitness/verdict/proposer lifecycle is LOG-ONLY — MEDIUM
[...]

Issue #1133, opened 2026-06-16
Trace correlation Phase 2: OTel span LINKS from agent.decision to the game trace (proper linkage, not just game.id attr)


aepfli:
Follow-up to #1088 Phase 1 (PR #1095, shipped: `game.id` correlation ATTRIBUTE on all agent spans). Phase 2 = proper OTel span **Links** so an `agent.decision` trace is navigable to the originating game-coordinator trace (not just filterable by a shared id).

## The gap
Agent spans are still standalone traces; `game.id` lets you *filter/join* in Jaeger but not *navigate* the trace tree. The agent has no parent context from the game (it consumes game state as METRICS, which carry no trace context).

## Approaches to evaluate + implement
1. **Metric exemplars (OTel-native, preferred IF feasible)**: game-coordinator attaches a trace exemplar (the game trace's `trace_id`) to per-game metrics; the agent extracts the exemplar `trace_id` during metric ingestion (`gamecontext/extract_metrics.go`) and adds an OTel span **Link** (not parent-child — the relationship is async/cross-trace) from `agent.decision` to that trace. **Investigate first** whether the OTLP metric pipeline (coordinator emit → collector → agent ingest) preserves exemplars and the Go SDK exposes them on ingest — if not cleanly supported, say so and use (2).
2. **Dedicated correlation signal**: coordinator emits the active game's `trace_id` as a metric attribute or a small per-game signal the agent reads; agent adds the span Link from it.

## Scope guidance (concurrency note)
- Agent side: the span-Link addition lives in the decision-span creation / `decision/telemetry.go` + ingest (`extract_metrics.go`) — disjoint from the intervention writer/ratelimit/prompt code.
- Coordinator side: the exemplar/trace_id emission goes at the per-game metric-emit call sites — keep it OUT of the base.py threshold/handicap logic (concurrent work there); prefer the metrics module / the emit lines.

## Acceptance
- An `agent.decision` span carries an OTel Link to the originating game trace (navigable in Jaeger), OR a documented finding that exemplars aren't feasible + the chosen alternative implemented.
- Falls back gracefully (no Link) when no game trace_id is available; no decision/gate behavior change.

Relates #1088 (Phase 1, #1095), #1095, #964.

Issue #1088, opened 2026-06-15
Connect agent decision spans/traces to their game's trace (currently orphan traces)


aepfli:
## Problem
Agent spans (`agent.decision`, `agent.llm.*`, `experiment.*`, `code_improvement.*`) are **standalone traces**, disconnected from the game they act on. In Jaeger you cannot pivot from a game ("show me what the agent decided/proposed for THIS game") to the agent's reasoning, or vice-versa. Confirmed on the live stack: an `agent.decision` trace has 3 spans tagged only `game.kind=shadow` — no `game.id` — and is a separate trace from the game-coordinator's per-game trace (`Joust Free-for-All`, `GameCoordinatorService/*`).

## Root cause
The agent ingests game state as **OTel metrics** (`gamecontext/extract_metrics.go`), which carry no trace context (no `traceparent`). So the agent has no parent span context from the game to attach to, and its decision spans start fresh root traces. The agent *does* know the `game.id` (metrics carry `game.id`/`game_id` per-data-point), it just isn't propagated onto spans or used to link.

## Proposed approach (two phases)
**Phase 1 — correlation attribute (cheap, high value):** stamp every agent span (`agent.decision`, `agent.evaluate`, `agent.llm.*`, `experiment.*`, `code_improvement.*`) with a queryable `game.id` (= the session_id `game_<uuid>` it already has). Then Jaeger/Grafana can filter agent spans by `game.id` and cross-reference the game-coordinator trace for the same id. Not one trace, but a reliable join key. (Today only `game.kind` is tagged.)

**Phase 2 — OTel span Links (proper causal linkage):** add `Link`s from `agent.decision` → the game's trace so they're navigable as related (Links, not parent-child — the relationship is asynchronous/cross-trace, which is exactly what Links are for). This needs the game's `trace_id` to reach the agent. Options:
- **Metric exemplars**: have game-coordinator attach trace exemplars to the per-player/per-game metrics; the agent extracts the exemplar `trace_id` and adds a span Link. (OTel-native, no new channel.)
- **Correlation signal**: game-coordinator emits the active game's `trace_id` as a metric attribute or a small dedicated signal the agent reads.
- Evaluate which is feasible with the current collector/SDK config.

## Acceptance
- Phase 1: from a game, filter Jaeger by `game.id=<id>` and see all agent activity for it; agent spans carry `game.id`.
- Phase 2: an `agent.decision` span Links to the originating game trace (navigable in Jaeger).

[...]

PR #1141, opened 2026-06-16
feat(agent): async LLM decision trace continuity via span Links (#1140 Slice A)

aepfli:
Part of #1088, #1140 (Slice A: async trace continuity)

## Problem
When inference fires async (#917), the evaluate cycle returns `nil` — no decision span in the `agent.signal_received` trace. The result lands later under a parent-less `agent.llm.apply` root, and the outbound HTTP call runs under another parent-less `agent.llm.infer.call` root. One logical decision was smeared across three disjoint traces.

## Fix
The #917 non-blocking design forbids parenting these into a shared trace, so this uses OTel span **Links** (cross-trace references), not parent-child — mirroring the #1133 `gameTraceLink` primitive.

- `decision.go`: at the async-fire site, capture the fire-cycle's currently-active SpanContext (`trace.SpanContextFromContext(ctx)`) and pass it into `fireInferAsync`. New local helper `fireCycleLink` builds a `trace.WithLinks` option (gated on `IsValid`) carrying `link.kind=fire_cycle` + `link.trace_id`.
- `async_infer.go`: thread the SpanContext (as a value, never a live ctx) through `fireInferAsync` -> `runInfer`, add the Link on `agent.llm.infer.call`, and stash it on `asyncOutcome.fireSC`.
- `async_apply.go`: add the Link on the `agent.llm.apply` root from `out.fireSC`.

Net effect: from the apply trace or the infer-call trace, a reader can navigate back to the originating `agent.signal_received` trace. Graceful no-op (no link, no error) when the fire-span SpanContext is invalid/missing — the production default today, since there is no otelgrpc interceptor and the signal_received root is only created on the sync decided path.

Observability only: decision logic, gating, and the non-blocking async behavior are unchanged. The #1133 `gameTraceLink` on the apply path's decision span and the `game.id` attributes are untouched.

## Tests
New `async_trace_continuity_test.go`: both `infer.call` + `llm.apply` carry a Link to the captured fire-cycle SpanContext when valid; absent (graceful) when no fire span; no panic; decision output unchanged (one dispatched action, `inference.applied=true`); plus a direct unit test of `fireCycleLink`.

`gofmt -l .` clean, `go build ./...`, `go vet ./...`, `go test ./decision/ -race -count=1` all green.

### DIFF
--- a/services/agent/decision/async_apply.go
+++ b/services/agent/decision/async_apply.go
@@ -47,7 +47,12 @@ func (l *Loop) applyAsyncResult(root context.Context, snapshot flags.Snapshot, t
 	// Export arrival) so the trace duration covers the whole fire->infer->apply span,
 	// the same backdating runDecision's root uses. It carries the game id so a Jaeger
 	// query by game.id finds the async decision alongside the sync ones.
-	applyCtx, applyRoot := l.Tracer.Start(root, SpanLLMApply,
+	// #1140 (Slice A): the apply root is parent-less by the #917 non-blocking design.
+	// Link it back to the fire cycle's originating agent.signal_received trace (carried
+	// as a SpanContext value on the outcome) so the apply trace — and the agent.decision/
+	// agent.action spans nested under it — is navigable from the signal that caused it.
+	// No-op when out.fireSC is invalid (the production default), exactly like gameTraceLink.
+	applyOpts := []trace.SpanStartOption{
 		trace.WithTimestamp(trig.T0),
 		trace.WithSpanKind(trace.SpanKindInternal),
 		trace.WithAttributes(
@@ -55,7 +60,9 @@ func (l *Loop) applyAsyncResult(root context.Context, snapshot flags.Snapshot, t
 			attribute.String(AttrGameID, l.gameID),
 			attribute.Int64(AttrLLMLatencyMs, out.latencyMs()),
 		),
-	)
+	}
+	applyOpts = append(applyOpts, fireCycleLink(out.fireSC)...)
+	applyCtx, applyRoot := l.Tracer.Start(root, SpanLLMApply, applyOpts...)
 	defer applyRoot.End()
 
 	// Emit the agent.llm.infer call span as a child of the apply root, with the REAL
--- a/services/agent/decision/async_infer.go
+++ b/services/agent/decision/async_infer.go
@@ -162,16 +162,30 @@ func (l *Loop) asyncEnabled() bool { return l.ctxProvider != nil }
 // APPLY path re-fetches the current context from the provider, never this snapshot —
 // that separation is the whole point of #917.
 //
-// #1140 (Slice A): fireSC is the fire-cycle's currently-active SpanContext, captured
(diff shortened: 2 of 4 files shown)


aepfli:
## Fix: anchor span at async-fire time (defect in original Slice A)

### The defect
The first cut captured `fireSC := trace.SpanContextFromContext(ctx)` at the async-fire site inside `decide()`. In production that `ctx` has **no active span**: `OnEvaluate` creates the `agent.signal_received` root only on the **sync decided** path (`len(decisions) > 0`, *after* `decide()` returns). An async-fire cycle returns `nil` from `decide()` (fireInferAsync returned true), so `OnEvaluate` hits the `len(decisions)==0` idle-return and creates **no span at all**. `fireSC` was therefore always invalid in production → the `fire_cycle` Links were a permanent no-op and the async-fire cycle had **zero trace presence**. The old test only passed because it injected an active span into the inbound ctx (`fireCycleCtx`).

### The fix
An async-fire cycle is **not idle** — it dispatched a decision, so it gets a real trace root. `fireInferAsync` now **creates the anchor `agent.signal_received` span itself**, immediately *after* the pile-up-guard CAS succeeds (a guard-skipped cycle dispatches no inference, so it produces no stray anchor). The anchor:
- reuses the **exact** sync-path shape — `SpanKindServer`, `trace.WithTimestamp(trig.T0)`, `rpc.system`/`rpc.service`/`rpc.method=Export`/`otlp.signal`/`session.id`/`game.id` — via a new shared `startSignalReceivedSpan` helper that the sync decided path and `emitDisabledSpan` now also call (their behavior unchanged — pure de-dup),
- is **ended immediately** (the async work runs later under its own linked roots; the anchor just represents "signal received → async inference dispatched"),
- yields an **always-valid** `SpanContext` that is threaded into the async job as before.

`agent.llm.infer.call` and `agent.llm.apply` Link back to that anchor unchanged (`link.kind=fire_cycle` + `link.trace_id` preserved). `async_apply.go` needed no change.

### Production topology now
- `agent.signal_received` (anchor root) — **now exists on async cycles**
- `agent.llm.infer.call` (root) — Links → anchor
- `agent.llm.apply` (root) — Links → anchor

### Tests
- **`TestAsyncContinuity_ProductionPathAnchored`** — drives `OnEvaluate` with a plain `context.Background()` (the production reality), asserts exactly one valid `agent.signal_received` anchor (Server kind, `game.id=g1`) exists **and** both async roots Link to *that anchor's* trace_id. This is the regression that would have caught the defect.
- **`TestAsyncContinuity_NoAnchorOnPileUpSkip`** — a guard-skipped cycle creates no anchor.
- **`TestAsyncContinuity_DecisionOutputUnchanged`** — output unchanged (1 dispatched action, `inference.applied=true`).
- **`TestFireCycleLink_Direct`** — keeps the invalid-`SpanContext` → no-op contract.

### Verification
`gofmt -l .` clean, `go build ./...` / `go vet ./...` clean, `go test ./decision/ -race -count=1` green, full `go test ./... -race` green. Files touched: `decision.go`, `async_infer.go`, `async_trace_continuity_test.go` only.

Issue #1143, opened 2026-06-16
auto_rubberband intervention — automatic skill-gap compression (#1103 Phase 3)


aepfli:
#1103 Phase-3 action. An AUTO (policy-style) intervention that dynamically rubber-bands difficulty to keep games close: detect the skill gap (leader vs laggard, from per-player intensity/skill in GameContext) and automatically apply a time-boxed handicap to compress it — strengthen the trailing player and/or ease pressure on them — reusing the set_player_handicap mechanism (#1107) and composing with partial_shield/soft_penalty (#1132/#1134) under the existing [0.5,2.0] clamp.

Unlike set_player_handicap (agent picks one player+factor), auto_rubberband is a single decision that the coordinator expands across players based on the live gap, bounded so it can never invert the standings or cause instant death. Shadow-only (append to shadow_experimental string variant). Value: off | gentle | strong (compression strength). MEDIUM weight, battery-gated.

Reuses: handicap composition, the #1133 patterns. See the #1103 design comment. Part of #1103.

aepfli:
Done in PR #1151 (squash-merged): auto_rubberband — headroom-ranked, bounded, time-boxed handicap boost to laggards (no-invert, double-clamped never-instant-death), MAX-composed, shadow-only with ci/agent.json parity. Part of #1103.

Issue #1134, opened 2026-06-16
Implement soft_penalty intervention (graduated alternative to permanent elimination) — #1103 Phase 3


aepfli:
#1103 Phase-3 action. A graduated, recoverable alternative to the all-or-nothing `eliminate_player`. Value `"warn"` (default) or `"tighten"`:
- **warn**: fire the existing `_warn_player` feedback (the visible flash+rumble) as an agent-driven "you're on notice" cue — recoverable, no state change.
- **tighten**: a short EXPIRING handicap REDUCTION (factor < 1, e.g. 0.6) for N seconds making that player temporarily EASIER to eliminate — pressure without guaranteeing it. This is the MIRROR of partial_shield (#1132 strengthens via max); tighten WEAKENS, so it needs a separate time-boxed mechanism (don't reuse partial_shield's max-only path; a low time-boxed factor that composes by min, or a distinct field). Bounded by the [0.5,2.0] clamp (can't make instant-death).

Shadow-only (append to the shadow_experimental STRING variant per #1127). MEDIUM weight (recoverable). Player-targeted, battery-gated. See the #1103 design comment.

Part of #1103.

aepfli:
Done in PR #1139 (squash-merged): soft_penalty graduated intervention — warn (recoverable cue) + tighten (time-boxed handicap reduction, MIN-after-MAX composition, double-clamped, shadow-only). Part of #1103.

Issue #1129, opened 2026-06-16
Implement partial_shield intervention (richer agent protection, reuses set_player_handicap) — #1103 Phase 2


aepfli:
#1103 Phase-2 action. Now that set_player_handicap (#1107) is merged + interventions actually APPLY (#1127 fixed), add `partial_shield`: an agent-specific protective action richer than the binary `grant_shield`. Instead of total immunity (skipping the death check entirely via grace_until), it applies a LARGE positive per-player handicap factor for N seconds — the player is much harder to eliminate but a genuinely huge spike can still take them out. Reuses the #1107 handicap-factor threshold mechanism (nearly free once that exists). Value `"<seconds>"` (or `"<seconds>:<boost>"`, boost default ~2.0 within the [0.5,2.0] clamp). Shadow-only (shadow_experimental allow-list). See the #1103 design comment.

Part of #1103.

aepfli:
Shipped in PR #1132: partial_shield intervention — timed per-player protection via max(handicap_factor, boost) under the [0.5,2.0] clamp (much harder to die but NOT immune, vs grant_shield's total grace_until immunity); only-strengthens, clean live expiry (no reset task), shadow-only (string variant per #1127). Reviewed/approved. #1103 Phase 2.

Issue #1107, opened 2026-06-15
Implement set_player_handicap intervention (multiplicative per-player threshold handicap, shadow-only) — #1103 MVP action 1


aepfli:
MVP action 1 from the #1103 design (maintainer chose: build both MVP actions; **multiplicative** handicap semantics reusing the existing clamp).

Add a `set_player_handicap` intervention: a per-player MULTIPLICATIVE factor on the death threshold, intent-framed ("help" = harder to eliminate / "rein in" = easier), distinct from and composing with `adjust_player_sensitivity`. Shadow-game-only initially (behind a new `shadow_experimental` interventions_allowed variant). Continuous rubber-banding — the replacement for the binary shield. See the #1103 design comment for full grounding (file:line).

Acceptance: new type_id + spec + handler + revert + writer case + shadow_experimental variant + weight (ratelimit.go + interventions.py) + one-line impact-model prompt fragment + tests; multiplicative, clamped [0.5,2.0]; default-safe (no game.json persistence); shadow-only.

Part of #1103.

aepfli:
Shipped in PR #1111: set_player_handicap intervention — per-player MULTIPLICATIVE death-threshold factor (clamped [0.5,2.0], composes with sensitivity, >1 harder/<1 easier), shadow-only via the shadow_experimental interventions_allowed variant (gate enforced on Go + Python sides). Reviewed/approved: math double-clamped safe, returns to neutral when untargeted, default-safe. #1103 MVP action 1.

PR #1151, opened 2026-06-16
feat(agent): auto_rubberband skill-gap-compression intervention (#1143)

aepfli:
Part of #1103, #1143

## What

`auto_rubberband` — a SINGLE agent decision (`off` | `gentle` | `strong`) that the **coordinator EXPANDS** across players from the live skill gap (#1103 Phase 3). Unlike `set_player_handicap` (agent hand-picks one serial+factor), the agent emits one decision and the game loop ranks players + applies the handicap.

## Expansion + compression logic (`base.py: apply_auto_rubberband`)

1. Rank alive players by **headroom** to their effective death threshold (`death - smoothed_accel`). Large headroom = comfortably leading; small headroom = on the edge (laggard). This is the live skill/standing proxy FFA already tracks (the EMA intensity feeding the death check).
2. For each player compute its `deficit` in `[0,1]` (laggard = 1.0, leader = 0.0) and a `target_boost = 1.0 + max_boost * deficit`.
3. Apply the boost as a time-boxed per-player handicap (`rubberband_until` / `rubberband_boost`), reusing the #1107 mechanism, composed by **MAX** alongside `partial_shield` in `_compute_effective_thresholds`.
4. Needs ≥2 alive players and a non-zero gap; otherwise a no-op.

## Bounds / clamp proof

- **No instant-death / no immunity**: the combined handicap is still clamped `[0.5, 2.0]` downstream, so even a maxed boost only raises the threshold ~2× (finite) — never immune (the deliberate distinction from `grant_shield`).
- **Per-strength cap**: the boost added above neutral 1.0 is capped — `gentle` +0.2, `strong` +0.4.
- **Never inverts standings**: standings = headroom to death, not the raw threshold. The boost is additionally capped so the boosted player's **resulting headroom can never exceed the leader's** (`death*boost - accel <= leader_headroom`) — compression toward the leader, never past it. The leader is never penalized.

## Composition precedence (`_compute_effective_thresholds`)

`handicap_factor` (#1107) → MAX with `partial_shield` boost (#1129) → MAX with `auto_rubberband` boost (#1143) → MIN with `soft_penalty` tighten (#1134) → clamp `[0.5, 2.0]`. Both strengthening boosts win by MAX; an active tighten still cuts through (pressure overrides protection).

## Value schema + malformed-safety

`off` (default / neutral / no-op) | `gentle` | `strong`. Validated in **both** layers:
- `writer.go: autoRubberbandOr` — unknown/malformed → neutral `"off"` no-op; **never escalates an unknown value to `"strong"`**.
- Python `parse_auto_rubberband_value` — neutral/unknown → `None` (safe skip).

## Shadow-only proof

Appended `,auto_rubberband` to the `shadow_experimental` `interventions_allowed` variant in `services/flagd/agent.json` only — absent from `ambient`/`standard`/`full`. The allow-list is the load-bearing gate. `TestAutoRubberbandIsShadowOnly` asserts presence in shadow + absence from all real variants. MEDIUM weight (`ratelimit.go`). No `game.json` persistence (bounded delta on the game object, expires with no reset task).

## Tests

- Go: `gofmt` clean, `go build`/`go vet` clean, `go test ./... -race -count=1` green (writer state+revert / default / malformed→off / case-insensitive; shadow-gate; weight via existing `costOf` parity; golden prompts regenerated with the new impact-model line).
- Python: full `game_coordinator` suite green (1180 passed). New coverage: gap-compression boosts the laggard not the leader; strong > gentle; per-strength cap; **never inverts standings** (headroom-based); never instant-death (finite threshold); expires; ≥2-player / equal-field / non-positive-seconds no-ops; extend-not-shorten/strengthen-not-weaken; MAX composition with partial_shield; soft_penalty MIN override; parser neutral/unknown→no-op; handler expand/neutral/malformed/no-game.

Main checkout clean (only the pre-existing dirty `flagd/ci/{agent,game}.json`, which were NOT touched).

### DIFF
--- a/services/agent/actions/testdata/interventions.json
+++ b/services/agent/actions/testdata/interventions.json
@@ -110,6 +110,15 @@
       },
       "defaultVariant": "none"
     },
+    "auto_rubberband": {
+      "state": "ENABLED",
+      "variants": {
+        "off": "off",
+        "gentle": "gentle",
+        "strong": "strong"
+      },
+      "defaultVariant": "off"
+    },
     "eliminate_player": {
       "state": "ENABLED",
       "variants": {
--- a/services/agent/actions/writer.go
+++ b/services/agent/actions/writer.go
@@ -54,6 +54,7 @@ const (
 	flagGlobalDifficultyFactor    = "global_difficulty_factor" // #766 F6
 	flagPacingProfile             = "pacing_profile"           // #766 F6
 	flagTempoRamp                 = "tempo_ramp"               // #1117 (#1103 MVP action 2)
+	flagAutoRubberband            = "auto_rubberband"          // #1143 (#1103 Phase 3)
 	flagEliminatePlayer           = "eliminate_player"
 	flagRevivePlayer              = "revive_player"
 	flagAudioCue                  = "audio_cue"
@@ -70,6 +71,7 @@ const activeVariant = "active"
 const (
 	neutralNone    = "none"
 	neutralDefault = "default"
+	neutralOff     = "off" // #1143 auto_rubberband neutral variant
 )
 
 // DefaultPath is the in-container interventions file path. Override with
@@ -121,6 +123,10 @@ const (
 	// gentlest, fully-recoverable "warn" — used when a Decision carries no Value or
 	// a malformed one. Never escalates an unparseable value to a tighten.
 	defaultSoftPenalty = "warn"
+	// defaultAutoRubberband is the safe fallback auto_rubberband strength (#1143):
+	// the gentler "gentle" — used when a Decision carries no Value. A malformed
+	// value degrades to "off" via autoRubberbandOr (never escalates to "strong").
+	defaultAutoRubberband = "gentle"
 )
 
 // partial_shield bounds (#1129) — mirror BaseGameMode's clamps so the writer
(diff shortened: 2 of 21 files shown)


aepfli:
Applied review fixes (commit 6ac7bec3):

1. **`services/flagd/ci/agent.json`** — added `auto_rubberband` to the `shadow_experimental` variant of `interventions_allowed`, restoring full parity with prod `agent.json`. Previously the dry-run/demo stack (which mounts `flagd/ci/`) had the flag *defined* (ci/interventions.json) but the allow-list **gate** denied it. The variant now ends `…partial_shield,soft_penalty,auto_rubberband`, byte-for-byte matching prod's `shadow_experimental`.

2. **`services/agent/decision/shadow_gate_test.go`** — `TestAutoRubberbandIsShadowOnly` now asserts presence-in-shadow + absence-from-real-variants on **BOTH** `flagd/agent.json` and `flagd/ci/agent.json` (same iterate-both-paths pattern as `TestSoftPenaltyIsShadowOnly` / `TestPartialShieldIsShadowOnly`). Removed the stale comment that rationalized skipping ci.

3. **`services/game_coordinator/lifecycle_handlers.py`** — corrected the `handle_auto_rubberband` docstring: the old text claimed battery gating happens inside the method's ranking, but `apply_auto_rubberband` reads no battery. Reworded to the actual rationale — rubberband only ever *raises* a threshold (benefit-only), so a low-battery laggard just gets a smaller benefit and is never penalized; no battery guard needed.

**Verified:** both flagd JSONs parse; `go test ./decision/ -race` passes incl. the restored both-paths shadow test; game_coordinator rubberband test files (167 tests) pass; gofmt clean, build + vet clean.

Issue #1168, opened 2026-06-16
agent.llm.prompt/retro spans carry the FULL uncapped system prompt every span — fingerprint it instead (root cause of the traces-pipeline collector OOM)


aepfli:
The collector OOM (#1167) is driven by SPAN SIZE on the traces pipeline, not span count or metrics (40-controller metric load is fine). The dominant payload: `agent.llm.prompt` (telemetry.go:186-191, llm_capture.go) and `agent.llm.retro` (retro_capture.go:56-57,441) attach the **FULL, UNCAPPED** prompt text:
- `llm.prompt.system` / `llm.retro.system` — the SYSTEM prompt: now LARGE (physics model #1091, mode fragment, impact magnitudes #1102, allow-list; retro got physics in #1160) and **essentially CONSTANT** — it's a pure function of low-cardinality `prompt_variant` + `mode` + `interventions_allowed`, all of which are ALREADY captured as span attributes.
- `llm.prompt.user` / `llm.retro.user` — the USER prompt: per-game/variable (the per-player arcs, timeline — the actually-interesting reasoning input).

Shipping the full constant system prompt on every llm span (4 shadow games/tick) is the redundant bulk. `llm.prompt.bytes` already records the size, so the text rides along for nothing.

## Fix (root cause, lets the collector stay lean)
- Replace the full SYSTEM-prompt text on spans with a FINGERPRINT: `llm.prompt.system_sha256` (short hash, e.g. 12 hex) + keep the existing `prompt_variant`/`mode`/allow-list attrs (the determinants). Do the same for `llm.retro.system`.
- Emit the full system-prompt TEXT exactly ONCE per distinct fingerprint for reproducibility — e.g. a low-rate dedicated `agent.llm.system_prompt` reference span/event (or a startup log) keyed by the hash, so any span's hash is resolvable to the exact text without paying per-span.
- KEEP the `user` prompt text (variable, the valuable reasoning input) — optionally cap at a generous bound, but don't fingerprint it.
- Keep `llm.prompt.bytes` (now ≈ user-prompt size).

Apply to BOTH the decision path (`llm_capture.go` / `telemetry.go` AttrLLMPromptSystem) and the retro path (`retro_capture.go` AttrLLMRetroSystem). This should shrink llm spans enough that the collector holds at the lean base ceiling — likely obviating the #1167 1.5Gi bump (keep base lean). Relates #1167, #1155, #1161, #739, #844, #1091.

[...]

Issue #1167, opened 2026-06-16
otel-collector still OOMs at 768Mi under sustained dry-run shadow-experiment span load (#1161 necessary but insufficient)


aepfli:
Found in verification run #7. #1161 fixed the memory_limiter/ceiling MISALIGNMENT (256→768Mi, limiter 600/spike-150) — necessary and correct — but 768Mi is still UNDER-PROVISIONED for the sustained dry-run shadow-experiment load: with the agent experiment loop running (4 paired shadow games/tick, ~20 in flight, each emitting decision/action/effect + experiment.* spans, all #1140 Links), the collector OOM-killed again (exit 137) within ~2 min, refused traces ("data refused due to high memory usage"), and dropped agent spans. Verification only captured spans in the window right after a fresh collector restart.

This is a CAPACITY/volume problem (distinct from #1161's alignment fix). Options:
- Raise the dry-run collector ceiling further (e.g. 1–1.5Gi) — simplest, ties to the stack-specific-envelope follow-up so the Pi base stays lean.
- Tail/probabilistic-sample SHADOW-game spans (the dominant producers) while keeping real-game + agent.decision/retro + experiment.verdict spans fully exported — preserves the demo-critical correlation traces, sheds the high-volume shadow noise.
- Reduce shadow-game span cardinality (do all per-shadow-game decision/effect spans need export?).

Prefer sampling shadow-game spans (highest leverage, keeps the demo traces intact) + a modest ceiling bump. Relates #1155, #1161, #1140, #828.

aepfli:
Done in PR #1170 (squash-merged): collector envelope scoped to experiment-heavy stacks — dry-run/gateway get 1.5Gi + limiter 1200/300 (live-validated to hold ~1.3GiB peak), Pi/prod base + ci stay lean 768Mi/600/150 via env-substituted limiter defaults. Immediate fix; the durable root-cause fix (constrain the filelog/logs pipeline, the real hog per verification #8) is #1173.

Issue #1091, opened 2026-06-15
LLM system prompt must explain the game's PHYSICAL mechanics + per-game-mode rules (LLM invents nonexistent mechanics)


aepfli:
## Problem (proven by the first real gemma4 output)
The agent's LLM system prompt frames the role/objectives/interventions but does NOT explain how the game physically works or how players interact. Evidence: gemma4's first real proposal justified raising `death_grace_period_seconds` *"to utilize downed opponents more actively, diversifying combat interactions beyond simple elimination focus."* In **FFA there are no downed opponents** — it's motion-controlled instant elimination, no respawn, no downed/dead-player state. The LLM invented a mechanic the game doesn't have, because the prompt never told it the physics.

## What the prompt is missing
1. **Physical model**: JoustMania is motion-controlled — each player holds a PS Move; the controller streams acceleration (intensity); moving too hard relative to the current threshold = elimination. Game speed / music tempo sets how much movement is allowed (tempo interpolates the effective death threshold — see #1082). There is a brief startup grace.
2. **Player interaction model**: players are NOT directly attacking each other; they compete on movement control vs the shared tempo/threshold. (Spell this out — the LLM assumed direct combat.)
3. **No dead-player utilization in FFA**: elimination is permanent; there is no respawn, downed state, or revive in FFA. (Respawn exists only in zombie/swapper modes.)

## Per-game-mode prompt (the bigger ask)
Each game mode works differently (FFA = last-mover-standing elimination; zombie/swapper = respawn/role-swap; tournament = bracket). The system prompt should include a **mode-specific block** explaining the rules and the valid levers for the *current* `game_mode`, so the LLM reasons within the actual mechanics instead of generic "combat." Build this as a per-mode prompt fragment selected by the game's mode, composed into the system prompt (alongside the existing objective/intervention sections in `services/agent/llm/prompt.go`).

## Scope notes
- Start with **FFA** (the only mode we test/run today) — get its physics + lever description right, validated against a real gemma4 proposal that no longer invents mechanics.
- Pair with #1082 (per-player movement + tempo context) — together they give the LLM both the *rules* and the *live state* to reason correctly.
- Companion to the dead-config finding (death_grace_period does nothing) — the LLM should also be told which levers actually have effect.

Relates #1082, #964, the prompt builder `services/agent/llm/prompt.go`.

aepfli:
[...]

PR #1169, opened 2026-06-16
fix(agent): fingerprint llm/retro system prompt instead of shipping full text

aepfli:
Part of #1168 (root-cause fix for #1167 collector OOM).

## What
The `agent.llm.prompt` (#739) and `agent.llm.retro` (#844) spans carried the FULL, uncapped system prompt text on every emission. That text is large and effectively constant for a given `prompt_variant` + `mode` + `interventions_allowed` (all already span attributes), so shipping it per span (4 shadow games/tick) was the dominant payload driving the #1167 traces-pipeline collector OOM — span SIZE, not count.

## Change
- **Fingerprint, not text**: `llm.prompt.system_sha256` / `llm.retro.system_sha256` (16 hex chars of SHA-256, via `systemPromptSHA`) now ride the span instead of the full system text.
- **Keep the user prompt** (the variable, valuable reasoning input — per-player arcs/timeline), capped at a generous 32 KiB (`capUserPrompt`) with a truncation marker so a pathological game cannot reintroduce per-span bloat.
- **`*.bytes` unchanged**: still `len(system)+len(user)` (full size), so the at-a-glance sizing number is the same; only the system TEXT stops being emitted.
- **Recoverability — once per fingerprint**: the full system text is emitted exactly once per distinct hash on a dedicated `agent.llm.system_prompt` reference **LOG** line (process-lifetime `sync.Map` of seen hashes, shared by both decision and retro paths) carrying `{system_sha256, prompt_variant, mode, interventions_allowed, system}`. **Log, not span**: the OOM is the traces pipeline — a reference span would add back to it; a log line keeps traces lean while any span's hash stays resolvable to its exact text.

## Scope
`services/agent/decision/{telemetry.go,llm_capture.go,retro_capture.go,decision.go}` + new `prompt_fingerprint.go` (hash helper, user-cap, once-per-fingerprint emitter) + tests. No collector config / experiment_loop / async / infracontext touched.

## Per-span reduction
~2.6 KiB+ system text → 16-byte hash per llm span (a minimal-snapshot floor; production prompts are larger with #1091/#1102/#1160).

## Verify
`gofmt -l . && go build ./... && go vet ./... && go test ./... -race -count=1` all green. Tests now assert: span carries the `system_sha256` hash and NOT the full system text (the large physics/allow-list strings are ABSENT); user prompt text still present; full system text emitted once per fingerprint (second identical cycle does NOT re-emit).

### DIFF
--- a/services/agent/README.md
+++ b/services/agent/README.md
@@ -740,18 +740,27 @@ The span carries the full prompt plus its attribution (single builder
 | `agent.prompt_variant` | resolved prompt variant |
 | `agent.objectives` | sorted `k=v` weights |
 | `interventions.allowed` | the allow-list summary |
-| `llm.prompt.system` / `llm.prompt.user` | the **full** prompt text (uncapped) |
+| `llm.prompt.system_sha256` | 16-hex **fingerprint** of the system prompt (#1168) — NOT the text; the full text is on the `agent.llm.system_prompt` log line |
+| `llm.prompt.user` | the user prompt text, capped at 32 KiB (a `...[truncated]` marker is appended if it exceeds the cap) |
 | `llm.prompt.bytes` | `len(system)+len(user)` |
 | `inference.configured` | the `model` flag |
 | `inference.used` | `"rules"` (the rules engine decided) |
 | `inference.fallback_reason` | `"no_backend_available"` |
 
 The companion log line `agent.llm.prompt_captured` carries only metadata
-(`session_id`, `variant`, `model`, `bytes`, `fallback_reason`) — the prompt text
-lives on the span alone. **View in Jaeger:** open `http://localhost:8080/jaeger/`,
-service `agent`, operation `agent.llm.prompt`. To replay a captured prompt against
-a real model, copy `llm.prompt.system` / `llm.prompt.user` into two files and run
-`scripts/replay-prompt.sh`. See
+(`session_id`, `variant`, `model`, `bytes`, `fallback_reason`). The **user** prompt
+text lives on the span; the **system** prompt text does **not** — to keep the traces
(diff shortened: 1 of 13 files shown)


aepfli:
Addressed the SHOULD-FIX review finding (replay docs/tooling pointed at the removed span attribute) in 9462d0c3:

- **`services/agent/README.md`** — prompt-capture (#739) and retro (#844) tables/text now reflect the fingerprint model: the span carries `llm.prompt.system_sha256` / `llm.retro.system_sha256` (16-hex fingerprint) plus the capped `llm.prompt.user` / `llm.retro.user`; the full system text lives on the once-per-fingerprint `agent.llm.system_prompt` log line (`{system_sha256, prompt_variant, mode, interventions_allowed, system}`). Replay path: resolve the span hash to that log line's `system` field, keep the user prompt from the span. Noted the system prompt is also deterministically reconstructable from (variant, mode, allow-list).
- **`services/agent/scripts/replay-prompt.sh`** — header how-to + usage now instruct pulling the user prompt off the span and the system text from `agent.llm.system_prompt` keyed by the span's `system_sha256` (`docker logs joustmania-agent | grep agent.llm.system_prompt | grep <hash>`). Retro path updated identically. (Script only echoes instructions — no programmatic attr extraction to repoint.)
- **`services/agent/gamewindow/render.go:14`** — comment now points at the `agent.llm.system_prompt` reference log's `system` field instead of the removed `llm.prompt.system` attribute.

Docs/comment only, no runtime logic touched. Skipped the capUserPrompt byte-boundary NIT as accepted.

Verification: `bash -n replay-prompt.sh` OK, `gofmt -l gamewindow/render.go` clean, `go build ./...` OK.

Issue #1182, opened 2026-06-16
Experiment-as-trace-root: child game spans via spawn-traceparent + analysis/verdict/code_improvement as experiment children (epic #1181 incr. 2)


aepfli:
Increment 2 of #1181. Make the experiment span the ROOT of its trace:
- `experiment` span = long-lived root. When the agent SPAWNS a shadow game, propagate the experiment's traceparent on the spawn call so the coordinator's game span is a CHILD (causal → parent-child, not a Link). Investigate the shadow-game spawn path (gamerunner / coordinator RPC) for where to inject context.
- In-game agent decisions/actions hang under each game span (composes with #1178).
- Per-game analysis (`agent.game.summary`, `experiment.fitness`, `agent.llm.retro`) = children of the EXPERIMENT span (timeline-correct: experiment outlives the ended game spans). Re-scopes #1142's experiment.* Links → experiment children; retro span → experiment child.
- Verdict + `code_improvement.promoted` = experiment's terminal children.
- Real (non-experiment) games keep the game-span-as-hub model (#1178) — no experiment parent. Graceful when no experiment context.
Part of #1181, #1088. Builds on #1178/#1175/#1164.

aepfli:
Done in PR #1196: shadow game spans → experiment children via spawn-traceparent; real games own-root (two guards, mutation-verified). Epic #1181 PR3.

Issue #1181, opened 2026-06-16
EPIC: make the experiment story legible — trace hierarchy + retro conclusions + an external experiment-dossier HTTP view


aepfli:
The agent's experiment is a HYPOTHESIS-TEST NARRATIVE: formulate a thesis → run trials (code-improvement #1011 / flag tweaks / in-game influence) → evaluate each game's outcome → refine or spawn more games → conclude (promote or discard). Today the pieces exist (cohort framework, journal #1025, fitness/verdict) but the STORY isn't legible. This epic makes it visible across two complementary surfaces.

## A. Trace hierarchy — experiment as the trace root (confirmed design)
- `experiment` span = root (long-lived). Shadow games it SPAWNS = CHILDREN (propagate the experiment's traceparent on the spawn call — causal, so parent-child not Link). In-game agent decisions/actions hang under each game span (#1178). Per-game analysis (summary, fitness, retro) = children of the EXPERIMENT span ("in between" games; the experiment outlives the ended game spans, so this is timeline-correct). Verdict + `code_improvement.promoted` = the experiment's terminal children.
- Real (non-experiment) games keep the game-span-as-hub model (#1178). Shadow/experiment games use this experiment-as-root model.
- Re-scopes: #1142 experiment.* Links → experiment children; retro span → experiment child; #1178 composes underneath.

## B. Retro actually concludes (#1179 — FIRST increment, in flight)
Wire the retro to call the LLM + capture its conclusion (assessment + suggestions). Today it's a no-op capture stub. The retro's suggestions become the THESES for the next experiment — closing the loop (the Proposer #960 is the link that turns retro analysis → experiment proposals).

## C. External experiment-dossier HTTP view (NEW — maintainer ask)
A SIMPLE HTTP endpoint on the agent serving the experiment narrative from the JOURNAL (#1025) — independent of Jaeger/Grafana:
- `GET /experiments` — list (id, thesis/candidate, status, verdict, started/concluded).
- `GET /experiments/{id}` — dossier: the THESIS (what idea/candidate/flag-tweak/influence is tested), arms, the games run with per-game fitness/outcome, the retro suggestions, the verdict (delta/significance/practical floor), the decision (promoted/discarded/refined), and deep-links to the game/experiment traces.
- JSON (and/or a minimal HTML view). This is the "what WAS this experiment" report a human reads.

## Increment order
1. **#1179** retro-conclusion (in flight) — needed regardless of hierarchy.
[...]

Issue #1179, opened 2026-06-16
Wire the retro to ACTUALLY call the LLM + capture its conclusion (assessment + suggestions) — agent.llm.retro is currently a no-op capture stub


aepfli:
The `agent.llm.retro` span captures the PROMPT but never calls the model — confirmed: a 10µs capture-only span (retro_capture.go:343-344 starts+ends it), `retroInference` (retro_capture.go:379-408) only resolves ATTRIBUTION (which tier would serve), never `Backend.Infer`. `inference.used=gemma4` is misleading. `llm/retro.go:20` documents it as a #739-style "capture-first" stub awaiting a backend-wiring follow-up that never happened (the in-game path got wired; retro didn't). There is also no retro-response decoder (`DecodeRetro`/`RetroResponse` don't exist). So the maintainer is blind: the agent's post-game LEARNING produces no conclusion because no LLM is asked.

## Fix (mirror the in-game capture pattern + #1168 fingerprint for raw text)
1. **Decoder** — add `llm.RetroResponse{ SessionAssessment string; Suggestions []RetroSuggestion; SessionFocus string }` + `RetroSuggestion{ InterventionType, Emphasis, Reason string }` + `llm.DecodeRetro(raw) (RetroResponse, error)` in `services/agent/llm/retro.go` (sibling of `llm.Decode`), defensive (tolerate fenced/prose-wrapped JSON), matching the contract at llm/retro.go:146-158.
2. **Call the backend** in `retro_capture.go OnGameEnd` — surface the live `res.Backend` from the resolver (currently computed in `retroInference` then discarded), and if non-nil call `backend.Infer(ctx, llm.Prompt{System, User})` under a `context.WithTimeout` budget, wrapping the outbound call in a client span (like `async_infer.go:275-287`) so #1112 traceparent propagates. If `res.Backend==nil` keep capture-only + `inference.fallback_reason=no_backend_available`.
   - **ASYNC**: OnGameEnd is a lifecycle hook — do NOT block it on a network call. Run the infer+decode+capture in a goroutine (mirror the in-game `runInfer`/`applyAsyncResult` split). The span ends after the response (real latency), not at 10µs.
[...]

PR #1185, opened 2026-06-16
feat(agent): wire retro to call the LLM + capture its conclusion (#1179)

aepfli:
Part of #1181, #1179.

`agent.llm.retro` was a 10µs no-op capture stub: it built the retrospective PROMPT and resolved which tier WOULD serve, but never called `Backend.Infer`, so the agent's post-game LEARNING produced no conclusion. This is increment 1 of epic #1181.

## What changed
- **Decoder** (`llm/retro.go`): `RetroResponse{SessionAssessment, Suggestions[], SessionFocus}` + `RetroSuggestion{InterventionType, Emphasis, Reason}` + `DecodeRetro(raw)` — a defensive parser (tolerates fenced / prose-wrapped JSON), sibling of the in-game `llm.Decode`. Recorded-only, so no required-field / vocabulary gating; the only hard requirement is "a single JSON object".
- **Async infer wiring** (`decision/retro_capture.go` `OnGameEnd`): surfaces the live `res.Backend` (previously resolved in `retroInference` then discarded) and, when non-nil, fires `infer+decode+capture` in a **goroutine** under a `context.WithTimeout` latency budget, wrapping the outbound call in an `agent.llm.infer.call` client span so #1112 traceparent propagates. The lifecycle hook never blocks; the span ends **after** the response (real latency). When `res.Backend==nil` it stays capture-only with `no_backend_available`.
- **Conclusion captured on the SAME span**: `llm.retro.session_assessment`, `llm.retro.session_focus`, `llm.retro.suggestion_count`, one `retro.suggestion` event per suggestion (intervention_type / emphasis / reason), `llm.retro.latency_ms`, `llm.retro.parse_ok`. On transport/timeout/decode failure it stamps `llm.infer.error` + an `agent.llm.retro_failed` WARN log.
- **Raw response once-log** (#1169 pattern): the reply rides `llm.retro.response_sha256` on the span; the full text is emitted **once per distinct hash** on the `agent.llm.retro_response` reference LOG line (keeps big text off the traces pipeline).
- New `llm.retro.*` attr keys in `telemetry.go`; raw-response reference singleton in `prompt_fingerprint.go` (sibling of `systemPromptRef`).
- `main.go` / `server.go`: `SetRootContext` + `AwaitInflight` so retro inference goroutines are bounded by the agent root context and joined on shutdown (no leak).

## Scope notes for review
- **Observability + a real inference call only** — no other behavior change. The budget gate already debited "one request"; now it matches reality.
- The retro span's **parenting is unchanged** (the experiment-as-root hierarchy is a separate later increment, #1182).
- Beyond the issue's strict trio (`retro.go` / `retro_capture.go` / `telemetry.go` + tests), this also touches `prompt_fingerprint.go` (the shared fingerprint infra, where the #1169 reference singleton belongs) and `main.go`/`server.go` (production wiring of the new SetRootContext/AwaitInflight seams). Did **not** touch `decision.go` / `async_*.go` / `effect.go`.
- **Please scrutinize the async / fail-open paths** (`runRetroInfer`): any flagd/backend/timeout/decode failure logs + stamps the error and ends the span cleanly — it must never break `OnGameEnd` or the game-end transition.

## Tests
- `DecodeRetro`: valid / fenced+prose-wrapped / empty-suggestions-valid / garbage+empty → typed errors.
- Retro span carries assessment/focus/suggestion_count + per-suggestion events on success; parse_ok=false + error on bad/garbage reply and on transport error; latency always recorded; raw-response once-per-hash; the goroutine does NOT block `OnGameEnd`; capture-only fallback when backend nil.
- `cd services/agent && gofmt -l . && go build ./... && go vet ./... && go test ./... -race -count=1` — all clean/green.

### DIFF
--- a/services/agent/decision/prompt_fingerprint.go
+++ b/services/agent/decision/prompt_fingerprint.go
@@ -62,6 +62,14 @@ func systemPromptSHA(system string) string {
 	return fmt.Sprintf("%x", sha256.Sum256([]byte(system)))[:systemPromptSHALen]
 }
 
+// responseSHA returns the short hex fingerprint of an LLM raw RESPONSE (#1169/#1179),
+// the same scheme systemPromptSHA uses for prompts: the first systemPromptSHALen hex
+// chars of the SHA-256 digest. It is stamped on the span as llm.retro.response_sha256
+// in place of the full raw text; the once-per-hash reference log resolves it.
+func responseSHA(raw string) string {
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(raw)))[:systemPromptSHALen]
+}
+
 // capUserPrompt bounds the user-prompt text emitted on a span to maxUserPromptBytes,
 // appending userPromptTruncatedMarker when it has to cut. The content is preserved
 // up to the bound — the user prompt is the variable reasoning input we keep (#1168),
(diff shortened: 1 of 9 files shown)


aepfli:
Applied the review fixes (commit ba1ef28b).

**SHOULD-FIX — root-cancellation now actually wired (`retro_capture.go`):**
The `root := rc.rootCtx` was computed and discarded; `inferCtx` derived only from the retro span ctx, so shutdown cancellation never reached an in-flight `Infer` — an in-flight retro would have waited out the full ~8s budget on shutdown, contradicting the comments here and in `server.go`.

Fix keeps both constraints:
- `inferCtx, cancel := context.WithTimeout(ctx, budget)` stays derived from the retro **span** ctx, so the `SpanLLMInferCall` client span still parents under the retro span (#1112 "one trace").
- Added a linked-cancel bridge:
  ```go
  if root := rc.rootCtx; root != nil {
      go func() {
          select {
          case <-root.Done():
              cancel()
          case <-inferCtx.Done():
          }
      }()
  }
  ```
  On shutdown `root.Done()` fires `cancel()`, promptly unblocking a well-behaved `Infer`; the goroutine also exits when `inferCtx` finishes normally, so it never leaks. (The in-game `async_infer.go` derives from `root` directly because it has no span-parent constraint — noted, not copied.)

**Corrected comments** in `retro_capture.go` (the `rootCtx` field, `SetRootContext` doc, and the goroutine itself) and in `server.go:154-157` to describe the real mechanism (root.Done() → cancel() → Infer returns promptly → AwaitInflight joins).

**NIT — `DecodeRetro` doc note (`llm/retro.go`):** added a contract caveat that the only guarantee is "a parseable JSON object"; a structurally-valid-but-empty object (`{"foo":1}`) yields `parse_ok=true` with an empty conclusion, indistinguishable from a genuinely empty healthy reply. No behavior change.

**New test** `TestRetro_RootContextCancellationUnblocksInfer` (the seam the reviewer flagged as untested): a ctx-honoring backend that blocks on `<-ctx.Done()`, a deliberately long 30s latency budget (so only the bridge — not the budget timeout — can unblock within the test window), then cancel `rootCtx`. Asserts `AwaitInflight` joins promptly (~10ms), the span ends stamped `parse_ok=false` + `llm.infer.error`, proving the call returned via cancellation rather than completing or leaking.

`gofmt -l .` clean, `go build ./...`, `go vet ./...`, `go test ./... -race -count=1` all green.

